PR #35, opened 2021-02-05
Docker: Update container for runtime

zarubaf:
Build a reasonable new CMake version as well as make `riscv32` available as a symlink to the multilib toolchain (bit of a hack)


fabianschuiki (approved):
Looks good to me!

PR #34, opened 2021-02-05
banshee: Implement AMOs

SamuelRiedel:
Add support for atomic operations. Currently supported are all RV32 atomics except LR/SC.

Todo:
- [ ] Implement LR/SC
- [x] Add unittests
- [x] Update changelog

PR #36, opened 2021-02-06
Banshee wfi

SamuelRiedel:
The RISC-V `wfi` instruction currently stops the simulation of a core. However, this operation is also commonly used to put cores to sleep, for example at barriers, or queue locks. To support this behavior I suggest the following:
- We have global variables to count how many cores are asleep and which cores should be woken up.
- Instead of terminating the thread immediately when getting a `wfi`, the thread instead checks if all cores are asleep. If this is the case, terminate the simulation.
- If some cores are still awake, the thread should yield. When it becomes active again, it should check for a wake-up signal or whether all cores are asleep.

Todo:
- [x] Discuss the behavior
- [x] Make sure the simulation is not ended prematurely. For example, if the only core awake triggers the next core in a queue lock and then goes to sleep itself, all cores are asleep, until the next thread sees its wakes-up trigger. We have to make sure none of the cores finish the simulation in such a situation.
- [x] Discuss the concept of time spent sleeping. Currently, we only count the retired instructions. The time a core spent sleeping is therefore always just the one `wfi` cycle. A possibility could be that the core waking up inherits the time of the core sending the trigger if this time is bigger. However, adding a notion of time should be discussed in a different PR, in my opinion.
- [x] Add unittest

This PR relies on #34.


SamuelRiedel:
@fabianschuiki What do you think?

sw/banshee/src/engine.rs
```diff
@@ -316,6 +316,12 @@ impl Engine {
             .map(|_| AtomicUsize::new(0))
             .collect();
 
+        // Allocate global variables to keep track of sleeping cores.
+        let num_sleep = AtomicUsize::new(0);
+        let wake_up: Vec<_> = (0..self.num_clusters * self.num_cores)
+            .map(|_| AtomicUsize::new(0))
+            .collect();
+
```
fabianschuiki, line 324: This seems like a nice way to track sleeping cores :+1:

sw/banshee/src/engine.rs
```diff
@@ -476,7 +492,20 @@ impl<'a, 'b> Cpu<'a, 'b> {
...
+                if value as i32 == -1 {
+                    self.num_sleep
+                        .fetch_sub(self.wake_up.len(), Ordering::Relaxed);
+                    for x in self.wake_up {
+                        x.fetch_add(1, Ordering::Relaxed);
+                    }
+                } else if (value as usize) < self.wake_up.len() {
+                    self.num_sleep.fetch_sub(1, Ordering::Relaxed);
+                    self.wake_up[value as usize].fetch_add(1, Ordering::Relaxed);
+                }
```
fabianschuiki, line 506: Neat!

sw/banshee/src/tran.rs
```diff
@@ -1903,7 +1903,21 @@ impl<'a> InstructionTranslator<'a> {
...
+                    LLVMCreateBasicBlockInContext(self.section.engine.context, NONAME);
+                let bb_wake_up = LLVMCreateBasicBlockInContext(self.section.engine.context, NONAME);
+                LLVMInsertExistingBasicBlockAfterInsertBlock(self.builder, bb_terminate);
+                LLVMInsertExistingBasicBlockAfterInsertBlock(self.builder, bb_wake_up);
+                LLVMBuildCondBr(self.builder, terminate, bb_terminate, bb_wake_up);
+                // Terminate
+                LLVMPositionBuilderAtEnd(self.builder, bb_terminate);
+                LLVMBuildRetVoid(self.builder);
+                // Continue
+                LLVMPositionBuilderAtEnd(self.builder, bb_wake_up);
```
fabianschuiki, line 1920: Much better 😉

sw/banshee/tests/wfi.c
```diff
@@ -0,0 +1,21 @@
...
+        *x = 1;
+        *wake_up = (core_id + 1) % core_num;
+    } else {
+        asm volatile("wfi");
+        *x += 1;
+        *wake_up = (core_id + 1) % core_num;
+    }
+    asm volatile("wfi");
+    return core_id == 0 ? core_num -= *x : 0;
+}
```
fabianschuiki, line 21: Nice test!

fabianschuiki (approved):
LGTM!

sw/banshee/src/engine.rs
SamuelRiedel: Thanks :)

sw/banshee/tests/wfi.c
SamuelRiedel: Perfect. Then we can merge :rocket:

PR #37, opened 2021-02-09
cluster_tb: Enable atomics on main memory

zarubaf:
That should make it possible to place LR/SCs and atomics on main (simulation memory).

I've needed to move the AXI Cut to after the atomic adapter, as I would otherwise get a deadlock. Briefly tracing this to the AMO module that seems to violate the AXI protocol on its downstream interface. To be checked in the future.


fabianschuiki (approved):
This looks very good to me, thanks! Should help a lot to develop software for the platform :-)

PR #38, opened 2021-02-12
snitch_cluster: Add wide slave port

zarubaf:
For Occamy (and other multi-cluster systems) we are going to finally need the wide slave port. Added by this PR.


fabianschuiki (approved):
Looks very good to me! Thanks for tackling this :+1:

PR #40, opened 2021-02-13
DMA bring-up

zarubaf:
One point for discussion could be a common prefix of the DMA runtime functions e.g., `snrt_dma...`. Otherwise, this is good to go from my side.


hw/ip/snitch_dma/src/axi_dma_tc_snitch_fe.sv
```diff
@@ -162,16 +163,8 @@ module axi_dma_tc_snitch_fe #(
...
-    ) i_axi_dma_tc_snitch_fe_id_gen (
-        .clk_i        ( clk_i                                          ),
-        .rst_ni       ( rst_ni                                         ),
-        .issue_i      ( twod_req_valid && twod_req_ready               ),
-        .retire_i     ( oned_trans_complete && twod_req_last_realigned ),
-        .next_o       ( next_id                                        ),
-        .completed_o  ( completed_id                                   )
-    );
+    `FFL(next_id, next_id + 'h1, twod_req_valid & twod_req_ready, 0)
+    `FFL(completed_id, completed_id + 'h1, oned_trans_complete & twod_req_last_realigned, 0)
```
fabianschuiki, line 167: 🤣 Love this change! Simplicity ❤️

hw/ip/snitch_dma/src/axi_dma_tc_snitch_fe.sv
```diff
@@ -241,125 +236,121 @@ module axi_dma_tc_snitch_fe #(
...
-        if(!rst_ni) begin
-            twod_req_q <= '0;
-        end else begin
-            twod_req_q <= twod_req_d;
-        end
-    end
+    `FF(twod_req_q, twod_req_d, '0)
 
-endmodule : axi_dma_tc_snitch_fe
+endmodule
```
fabianschuiki, line 356: Nice refactoring!

sw/cmake/toolchain-gcc.cmake
```diff
@@ -5,7 +5,7 @@ set(CMAKE_OBJDUMP riscv32-unknown-elf-objdump)
 set(CMAKE_INTERPROCEDURAL_OPTIMIZATION true)
 
 add_compile_options(-march=rv32imafd -mabi=ilp32d -mcmodel=medany -mno-fdiv -ffast-math -fno-builtin-printf -fno-common)
-add_link_options(-march=rv32imafd -mabi=ilp32d -nostdlib -Wl,-Ttext-segment=0x80000000)
+add_link_options(-march=rv32imafd -mabi=ilp32d -nostartfiles -Wl,-Ttext-segment=0x80000000)
```
fabianschuiki, line 8: Good call. No need to start implementing these functions by hand again 😄

sw/snRuntime/src/memcpy.c
```diff
@@ -2,32 +2,9 @@
...
-    } else {
-        while (dst != end) {
-            *(char *)dst++ = *(char *)src++;
-        }
-    }
-
-    return dst;
-}
+// TODO: Implement using cluster DMA for a faster `memcpy`.
+void *snrt_memcpy(void *dst, const void *src, size_t n) { memcpy(dst, src, n); }
```
fabianschuiki, line 10: Nice!

sw/snRuntime/include/dma.h
```diff
@@ -0,0 +1,158 @@
...
+               (  0b00000 << 20) | \
+               (     (14) << 15) | \
+               (    0b000 << 12) | \
+               (     (10) <<  7) | \
+               (0b0101011 <<  0)   \n"
+        : "=r"(reg_txid)
+        : "r"(reg_size));
+
+    return reg_txid;
+}
```
fabianschuiki, line 50: My suggestion would be to move all of these functions into a `src/dma.c` file and rely on link-time optimization to inline them at the very end. Otherwise it becomes increasingly difficult to swap in different implementations of the runtime depending on the chosen system (Until now the header files are system-invariant, which would be a nice thing to keep.)

Also, how about we prefix all these functions with `snrt_`, e.g., `snrt_dma_start_1d`, and move all the function prototypes to the single `snrt.h` header? This will keep the include space clean.

Just some nitpicking. What do you think @zarubaf ?

sw/snRuntime/tests/dma_simple.c
```diff
@@ -0,0 +1,43 @@
+// Copyright 2020 ETH Zurich and University of Bologna.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include <dma.h>
```
fabianschuiki, line 5: This include would be subsumed by the include of `snrt.h` if we go with the above suggestion.

sw/snRuntime/include/dma.h
zarubaf: No nitpicking at all - this is useful advice and I agree. I've changed the implementation so that the prototypes are now in `snrt.h` and the definitions are in a separate compilation unit (`dma.c`). Agree that this looks much cleaner. I've also seen that LTO is pretty aggressive so I am very positive about this change!

fabianschuiki (approved):

PR #41, opened 2021-02-15
Add separate imm/reg SCFG instructions and matmul benchmark

fabianschuiki:
Rename the current SCFGR and SCFGW instructions to SCFGRI and SCFGWI. Add a new implementation of SCFGR and SCFGW which reads the register and data mover indices from a register. This allows for code without extensive compiler support to read or write SSR configuration registers.

Also add runtime support for SSRs and a first matrix multiplication benchmark application.


hw/ip/snitch_cluster/src/snitch_cc.sv
```diff
@@ -587,10 +587,35 @@ module snitch_cc #(
     `FF(cfg_req_valid_q, cfg_req_valid, 0)
     `FF(cfg_rsp.id, ssr_cfg_req.id, 0)
 
-    assign ssr_cfg_req.id = acc_snitch_demux_q.id;
-    assign ssr_cfg_req.word = acc_snitch_demux_q.data_op[31:20];
-    assign ssr_cfg_req.write = acc_snitch_demux_q.data_op ==? riscv_instr::SCFGW;
-    assign ssr_cfg_req.data = acc_snitch_demux_q.data_argb[31:0];
+    always_comb begin
+      import riscv_instr::*;
```
zarubaf, line 591: Didn't know that this works :-) Guess that is the advantage of a standard where "everything goes"

zarubaf (approved):
LGTM

PR #47, opened 2021-02-23
snitch_cluster: Add missing include in Bender file

noah95:
Not sure about the correct fix but this helps me compile for modelsim

```
make bin/snitch_cluster.vsim
```

Else, the `axi/typedef.svh` is not foud

### DIFF
--- a/hw/system/snitch_cluster/Bender.yml
+++ b/hw/system/snitch_cluster/Bender.yml
@@ -14,9 +14,14 @@ dependencies:
 
 sources:
 # Level 0:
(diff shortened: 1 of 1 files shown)


zarubaf:
That shouldn't be necessary as the cluster already depends on `axi` which exports the include location. What problem are you running into?

noah95:
I got a `File not found` from `vlog` when compiling `snitch_cluster_wrapper.sv`, but after cleaning and recompiling it works without the changes proposed. Probably an artifact from a previous error. Closing

PR #49, opened 2021-02-23
sw: Fix cluster software cmake cache variables

fabianschuiki:
Mark the variables declared by the `snitch_cluster` system software CMakeLists.txt as `CACHE`, to then be picked-up appropriately by the `SnitchUtilities.cmake` file later on. Otherwise the wrong simulator and runtime would be used.


zarubaf (approved):
Awesome. Nasty build environment thing :-D

fabianschuiki:
Indeed. Let's see if the cluster CI passes with this.

PR #50, opened 2021-02-23
Improve SSR-only matmul benchmark

fabianschuiki:
Apply some loop unrolling also there, such that the SSR-only implementation of matmul is comparable to the baseline one.


zarubaf (approved):
Awesome. LGTM.

PR #52, opened 2021-02-23
Open source Thestral solder.py

zarubaf:
Solder was very useful for the Thestral tape-out so here is my attempt to make it a bit more general (removing some Thestral dependencies) and using it for Occamy.

This will also be useful when open-sourcing Thestral :-)


fabianschuiki (approved):
Excellent idea to open source this!

PR #53, opened 2021-02-24
banshee: Fix lifetime of CString pointer

SamuelRiedel:
Fix the lifetime of a pointer to avoid access to a potentially deallocated CString.


fabianschuiki (approved):

fabianschuiki:
Seems like the master is already broken. Let's just merge this nice fix anyway 👍 💯

PR #54, opened 2021-02-24
banshee: Add support for fmv.x.w and fmv.w.x

noah95:
This should implement the `fmv.x.w` and `fmv.w.x` instructions. Not sure if it's correct, open for criticism ;)

### DIFF
--- a/sw/banshee/src/tran.rs
+++ b/sw/banshee/src/tran.rs
@@ -1658,6 +1658,40 @@ impl<'a> InstructionTranslator<'a> {
         trace!("{} x{} = x{}", data.op, data.rd, data.rs1);
         let name = format!("{}\0", data.op);
         let _name = name.as_ptr() as *const _;
+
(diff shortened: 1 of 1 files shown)


fabianschuiki:
I think this doesn't do what you want at the moment. The implementation now reads the integer register `rs1` and stores it to the integer register `rd`. You probably want to read from integer `rs1` and store to *float* `rd`, and vice versa for the other instruction :-) My suggestion would be to add a separate `match` above the current stuff in the function and only check for the `fmv.*.*` operations there. This way you can carefully use `read_freg`, `read_reg`, `write_freg`, and `write_reg` as appropriate (which will help keep the trace clean :-) ).

noah95:
Oh of course, I had no idea what I was doing. Will look into it

fabianschuiki:
Thanks a lot for tackling this! 👍

PR #55, opened 2021-02-25
banshee: Add support for fmv instructions

noah95:
Next attempt!

This implements `fmv.x.w` as a float store and `fmv.w.x` as an integer store. Both work as tested in the newly created test `fmv.c` which was compiled with llvm.

Open points
 - [x] Do I have to specify how the new unit test is built? Or are the source/dump/binary simply checked in
 - [x] Did I understand the behavior of `fmv` correctly? Are the upper 32-bit of the floating point register left as-is or modified, and how is the behavior of an `LLVMBuildStore` in this case?


sw/banshee/src/tran.rs
```diff
@@ -1658,6 +1658,51 @@ impl<'a> InstructionTranslator<'a> {
...
+                self.trace_access(
+                    TraceAccess::WriteFReg(data.rd as u8),
+                    LLVMBuildLoad(self.builder, raw_ptr, NONAME),
+                );
+                return Ok(());
+            }
+            _ => (),
+        }
+
+        // Handle DMA operations
```
fabianschuiki, line 1705: Nice solution!

sw/banshee/tests/fmv.c
```diff
@@ -0,0 +1,21 @@
...
+        "fmv.w.x %1, x0\n"  // float = 0
+        "fmv.w.x %1, %0\n"  // float <- int
+
+        "mv      %0, x0\n"  // int   = 0
+        "fmv.x.w %0, %1\n"  // int   <- float
+
+        : "+r"(d_rd), "+f"(f_rd));
+
+    return 0;
+}
```
fabianschuiki, line 21: Neat!

fabianschuiki (approved):
Looks very nice to me! We've decided to check in the compile RISC-V binaries for unit and integration tests to make it much simpler to cross-check modifications to Banshee without needing a full compiler stack all setup. So :+1: for including them here :smiley: .

fabianschuiki:
Ready to merge when you are :+1:

fabianschuiki:
One minor comment: You might want to add the fmv test to `sw/banshee/tests/Makefile`, just in case :smiley:

noah95:
> One minor comment: You might want to add the fmv test to `sw/banshee/tests/Makefile`, just in case smiley

Done. I will look into the behavior of the upper 32-bits later today

noah95:
from rtl simulations, this does now do the same so can be merged imo

PR #58, opened 2021-02-26
banshee: Consider the hartid offset in the `wfi` implementation

SamuelRiedel:
The `wfi` vector storing which core has a pending wake-up request did not consider the `base-hartid`. This led to out-of-bound array accesses if `base-hartid > 0`.

This should fix the failing CI.


SamuelRiedel:
The _SW on banshee (Container)_ test still fails, I assume we just have to rerun the tests to use the updated container? And from then on, no failing container should make it into master?

zarubaf:
Yes, exactly. After we merged this into master everything should work again. Thanks!

PR #59, opened 2021-02-26
Banshee Cycles

SamuelRiedel:
Add the capability to track the latency of each instruction.


fabianschuiki (approved):
Very nice addition, thanks a lot! :+1:

PR #63, opened 2021-03-04
Add more fcvt float convert instructions

noah95:
Specifically
```
fcvt.w.d
fcvt.w.s
fcvt.wu.s
fcvt.wu.d
fcvt.d.s
fcvt.s.d
```

And added tests for all currently supported float conversions.

The Test takes a long time to compile, most time is spent in `Optimizing module` and `Creating JIT compiler for translated code`. Perhaps @zarubaf or @fabianschuiki could look into that?


SamuelRiedel:
Very nice, thanks a lot. 👍

PR #64, opened 2021-03-04
banshee: Allocate FREP max repetition count to prevent IR use domination

paulsc96:
This fixes a Banshee binary translation issue regarding FREP wherein the generated LLVM IR may (through branches or jumps into a loop body) access an SSA value before assignment, leading to an corresponding use domination error on translation.


SamuelRiedel:
LGTM. Thanks :+1:

PR #69, opened 2021-03-18
Skip the cache line currently used in the eviction policy

SamuelRiedel:
This PR addresses the implicit assumption in Snitch, that the instruction interface has to be stable once the instruction is ready. Snitch will violate the protocol at the output interface otherwise. 

For example, when we execute a load but the memory interface is not ready, Snitch will already have enabled the valid signal and has to hold this signal until it gets a ready. However, the current cache policy might replace the cache line holding this load instruction with the prefetched line, which would lead to the cache withdrawing the ready signal. This leads to Snitch violating the protocol on the memory interface.

This PR addresses this issue with:
- Updated cache eviction policy to make the interface stable.
- Added assertion in Snitch, checking for a stable instruction interface.

A fall-through register at the interface would be another option, but that would impact the critical path.

### DIFF
--- a/hw/ip/snitch/src/snitch.sv
+++ b/hw/ip/snitch/src/snitch.sv
@@ -6,6 +6,7 @@
 // Description: Top-Level of Snitch Integer Core RV32E
 
 `include "common_cells/registers.svh"
+`include "common_cells/assertions.svh"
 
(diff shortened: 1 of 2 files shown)


zarubaf (approved):
This is terrific! Thanks a lot for this fix!

PR #70, opened 2021-03-18
ssr: Unpacking of SSR address according to ISA documentation

noah95:
Configuring the SSR with the `scfgwi` instructions as [described in the documentation](https://github.com/pulp-platform/snitch/blob/master/docs/rm/custom_instructions.md#configuration-register-operations) is not working due to a wrong unpacking of `dm` and `reg` fields.

With these changes, the `scfgwi` instruction is correctly decoded according to this table

| imm[11:4] | imm[4:0] | rs1   | funct3 | rd    | opcode     | operation |
|:---------:|:--------:|:-----:|:------:|:-----:|:----------:|:---------:|
| 7         | 5        | 5     | 3      | 5     | 7          |           |
| reg       | ssr      | value | 010    | 00000 | OP-CUSTOM1 | SCFGWI    |

The immediate read `scfgri` doesn't work either. The problem here is that the `mem_resp_valid_i` signal is asserted a clock cycle delayed on the memory side of `i_stream_to_mem`. But this is topic of a new pr.

### DIFF
--- a/hw/ip/snitch_cluster/src/snitch_cc.sv
+++ b/hw/ip/snitch_cluster/src/snitch_cc.sv
@@ -591,7 +591,7 @@ module snitch_cc #(
       import riscv_instr::*;
       automatic logic [11:0] addr;
       automatic logic [4:0] addr_dm;
-      automatic logic [4:0] addr_reg;
+      automatic logic [6:0] addr_reg;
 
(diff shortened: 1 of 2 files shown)


fabianschuiki (approved):
I like the more explicit nature! The concatenations on the left-hand side of `=` were admittedly a bit hard to reason about :wink:

PR #74, opened 2021-03-23
ssr: Move to standalone IP and create `snitch_ssr` module

paulsc96:
This change should simplify future work on SSRs, as well as standalone testing, by
* providing an own IP directory for them instead of including them in `hw/ip/snitch_cluster`
* separating SSR address generator and data mover logic into a standalone `snitch_ssr` module.

Note that these changes do not introduce any functional changes and should not cause any regressions. They are intended to simplify understanding of SSRs and later SSR extension and experimentation.


zarubaf (approved):
Great, thanks - that looks very good to me!

PR #76, opened 2021-03-24
banshee: Check if return value was written at all

SamuelRiedel:
The `exit_code` register is initialized to 0 in Banshee. If Banshee stops before writing a return value, it will look as if the execution was successful. This patch checks the LSB of the `exit_code` to make sure the return value was actually written. If we have debug prints enabled, it will generate a warning, otherwise, it will just set Banshee's return value to 1.

### DIFF
--- a/sw/banshee/src/engine.rs
+++ b/sw/banshee/src/engine.rs
@@ -370,10 +370,13 @@ impl Engine {
 
         // Print some final statistics.
         trace!("Final state hart {}: {:#?}", cpus[0].hartid, cpus[0].state);
-        info!(
-            "Exit code is 0x{:x}",
-            self.exit_code.load(Ordering::SeqCst) >> 1
-        );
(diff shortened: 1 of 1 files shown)


paulsc96 (approved):
Good catch! My only suggestion is to maybe choose a return code that will stand out more, but that is arguable.

SamuelRiedel:
Yes, I thought about that as well, but I did not know what would be a good return value. Maybe `0xDEADBEEF` or `0xBADCAB1E`, but this would get lost since the shell uses decimal. Maybe just something like `1234`?

paulsc96:
I think most shells (including bash) truncate to unsigned bytes, so 0-255 is viable.

To give it a Swiss touch, how about 117, since this is illegal behavior :)

SamuelRiedel:
I like it. :+1:

PR #78, opened 2021-03-24
banshee: Fix return value checking

SamuelRiedel:
A stupid mistake made it into the previous merge, which inverted the check for the return value. This is fixed in this PR.

Please do not merge yet, I am also fixing some binaries which currently fail.


SamuelRiedel:
Ok, next try. I fixed the `eoc` computation and the binaries which didn't have a return value.

PR #79, opened 2021-03-24
tcdm-interface: Do not assert stable write data and strobe on reads

paulsc96:
In their current implementation, SSRs may violate the assertion in `hw/ip/tcdm_interface/tcdm_interface.sv:108` when their IO timings are nonzero.

This assertion checks that the request TCDM payload remains stable as long as it is valid, but the downstream is not yet ready. This includes write data, even if the request is a read.

However, SSRs directly connect the request write data to the bidirectional FIFO output and assert a request valid as long as an address is ready and there are credits. Since the cycle delay at which the FIFO changes its head contents is decoupled from the application delay of the ready signal at the IO interface, the lock-in assertion may be violated.

This PR masks the write data to `'0` on read, ensuring request stability on reads and writes with some logic overhead.

### DIFF
--- a/hw/ip/tcdm_interface/src/tcdm_interface.sv
+++ b/hw/ip/tcdm_interface/src/tcdm_interface.sv
@@ -105,8 +105,8 @@ interface TCDM_BUS_DV #(
   assert property (@(posedge clk_i) (q_valid && !q_ready |=> $stable(q_addr)));
   assert property (@(posedge clk_i) (q_valid && !q_ready |=> $stable(q_write)));
   assert property (@(posedge clk_i) (q_valid && !q_ready |=> $stable(q_amo)));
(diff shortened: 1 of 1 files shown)


zarubaf:
Thanks for tackling this.

To be sure to understand this: Adapting the assertion to only check stability on the write data when there is actually a write going on has the same effect, right? If so I would much more prefer this route.

I don’t see any direct advantage of silencing this - afaik the SSR to memory path is already quite critical in terms of timing and we should save any logic we can. Do you agree? Is there an advantage I don’t see?

The compliance to OBI imho could be done in an adapter.

paulsc96:
This is correct. I adapted the assertion in `tcdm_interface.sv` to only require write data and strobe stability on reads and removed the write masking.

Is this better?

zarubaf:
Perfect, thanks a lot!

PR #80, opened 2021-03-24
ssr: Standalone fixture and testbench for out-of-context verification

paulsc96:
This PR adds a fixture and a simple, extensible testbench for SSRs, along with some test data. The current test sequence is fairly basic, but can be extended easily.

The SSR fixture can be a helpful starting point to integrate and verify future extensions for SSRs while avoiding regressions.


hw/ip/snitch_ssr/test/future/tcdm_test_extra.sv
```diff
@@ -0,0 +1,140 @@
+// Copyright 2020 ETH Zurich and University of Bologna.
+// Solderpad Hardware License, Version 0.51, see LICENSE for details.
+// SPDX-License-Identifier: SHL-0.51
+
+/// A set of testbench utlities for the TCDM interface.
+/// This package proposes an additional driver class using non-random structs as arguments.
+/// TODO: Integrate this functionality properly in tcdm_test.
```
zarubaf, line 7: Why can't you use the `rand` version? If you do not call on `randomize` that should be equivalent, shouldn't it?

zarubaf, line 7: Why can't you use the `rand` version? If you do not call on `randomize` that should be equivalent, shouldn't it?

zarubaf (commented):
Looks good, would be nice to get rid of the code duplication with the `rand` vs non-`rand` version.

hw/ip/snitch_ssr/test/future/tcdm_test_extra.sv
paulsc96: I was under the impression that since the request and response classes had no explicit constructor and the fields are not accessible, I could not create user-defined requests and responses. I will double-check and see If I can use the existing package.

paulsc96: I was wrong :sweat_smile: By default, SV class fields are public, so I can use the existing classes just fine.

_But_ there is an oversight in the classes in `tcdm_test`: they do not pass on parameterization to the `req_t` and `rsp_t` types they reference, leading to errors when using nonstandard parameters.

I fixed this now. Should I commit the fix with this PR or make a separate one?

zarubaf: Perfect, thanks. Yes, I messed up the parameterization. Thanks for fixing it. This PR is just fine.

zarubaf:
Ready from your side?

zarubaf (approved):

paulsc96:
Yes, ready :)

PR #82, opened 2021-03-25
Make Banshee configurable

SamuelRiedel:
I reworked the whole configuration to work as follows now:

- Configuration related structs and functions are stored in [configuration.rs](sw/banshee/src/configuration.rs)
- We have a few hierarchical structs that describe the different aspects of the architecture.
- All these structs have default values that match the current system. To check or modify the system parameters, only this file has to be changed.
- Nothing changes for people using the default configuration. So you do not have to use the `--configuration` flag to continue working with the current system.
- We provide functions to parse a JSON or YAML configuration file into this struct.  
- We also provide a function to generate a default configuration file. You can use it to reference which parameters can be changed and how the file should look like. It contains the default configuration. To generate it, you can use: `banshee --create-configuration default.json/yaml`. It will write the full default configuration in JSON or YAML format. See below for the current default config.

What can we configure:
Some parameters like the memory hierarchy or the available control registers are 'hardcoded' since they require the engine and translation stage's corresponding infrastructure. However, their addresses or latency are configurable.
- All the addresses of the control registers or other special addresses like the UART
- The start and end address of the TCDM/DRAM and their latency
- The latency of every instruction
- If you want to specify one control register, you have to specify all of them. This can be improved in the future, but we also have to be careful that the default config does not map two control registers to the same address in such a case.
- You can specify just a subset of instructions; the rest will have a latency of 1 cycle.
If you modify the memory map, you have to specify the TCDM and DRAM's full configuration. Again, this is just a limitation of the current parsing function.

The default configuration in YAML format as generated by Banshee. Note, YAML does support hex, so you can write the addresses in hex, but I didn't manage to write hex values without reverting to using strings instead of numbers.
```YAML
---
address:
  barrier_reg: 1073741880
  cluster_base_hartid: 1073741888
  cluster_id: 1073741904
  cluster_num: 1073741896
  nr_cores: 1073741840
  scratch_reg: 1073741856
  tcdm_end: 1073741832
  tcdm_start: 1073741824
  uart: 4027285504
  wakeup_reg: 1073741864
inst_latency:
  add: 1
  addi: 1
# ...
  xor: 1
  xori: 1
memory:
  dram:
    end: 2415919104
    latency: 10
    start: 2147483648
  tcdm:
    end: 131072
    latency: 2
    start: 0
```

---

Previous version:

This PR aims to remove all hardcoded architecture parameters from Banshee and instead read them from a configuration file. This will lead to a generic Banshee binary that is able to simulate multiple platforms. The configuration can be set via the command line. E.g. `banshee --config manticore.json hello_world`

This is very much still a work in progress, but I would like to get your opinions on how to tackle the configuration.

1. [Manticore.json](https://github.com/SamuelRiedel/snitch/blob/71f1f0f281d21a014242a99d70ca8dbb9429005a/sw/banshee/manticore.json) is an example of how to configure Banshee. You can set the TCDM location and size, the control registers, and the latencies of the loads to different regions and each instruction. Is there anything else you would like to configure in the near future? Any suggestions on the format? For example, we could clearly split the memory latencies from the instruction latencies and instead specify the latency with the memory range.
2. How should we handle defaults? The current implementation requires a JSON configuration file, otherwise, the control register addresses are not set. This would break the previous flow because we have a new mandatory command line flag. We could either parse the `manticore.json` by default, but that prevents the binary from being relocatable. Or, we add a new struct in Rust that represents the default configuration, and we only parse the JSON to overwrite some values. I prefer the latter because it will not break anyone's flow and we have the default configuration as a 'documentation' of what can be set via JSON. However, this also means that we need to create a default configuration. The latency for each instruction can be [generated](https://github.com/SamuelRiedel/snitch/blob/71f1f0f281d21a014242a99d70ca8dbb9429005a/sw/banshee/src/riscv.rs#L2595) quite easily with `riscv-opcodes`. I suggest writing the architecture configuration by hand as a Rust struct to avoid too many levels of indirection and generates. What do you think?
3. I will also add other command line options, such as `num_cores` or `num_clusters`, to the config. Do you agree that this would be nice to have? The order of precedence would then be:
      - Specific command line switch. I.e.: `--num_cores`
      - Config from JSON file
      - Default configuration


SamuelRiedel:
@fabianschuiki  @zarubaf @paulsc96 Can you quickly let me know what kind of architecture parameters you would like to have configurable and how we should handle a default configuration?

zarubaf:
That is really, really nice @SamuelRiedel!
> 
> 1. [Manticore.json](https://github.com/SamuelRiedel/snitch/blob/71f1f0f281d21a014242a99d70ca8dbb9429005a/sw/banshee/manticore.json) is an example of how to configure Banshee. You can set the TCDM location and size, the control registers, and the latencies of the loads to different regions and each instruction. Is there anything else you would like to configure in the near future? Any suggestions on the format? For example, we could clearly split the memory latencies from the instruction latencies and instead specify the latency with the memory range.

Small nitpick. Can we re-name this `occamy.json`? Then we can further develop this together with the RTL in here.

> 2. How should we handle defaults? The current implementation requires a JSON configuration file, otherwise, the control register addresses are not set. This would break the previous flow because we have a new mandatory command line flag. We could either parse the `manticore.json` by default, but that prevents the binary from being relocatable. Or, we add a new struct in Rust that represents the default configuration, and we only parse the JSON to overwrite some values. I prefer the latter because it will not break anyone's flow and we have the default configuration as a 'documentation' of what can be set via JSON. However, this also means that we need to create a default configuration. The latency for each instruction can be [generated](https://github.com/SamuelRiedel/snitch/blob/71f1f0f281d21a014242a99d70ca8dbb9429005a/sw/banshee/src/riscv.rs#L2595) quite easily with `riscv-opcodes`. I suggest writing the architecture configuration by hand as a Rust struct to avoid too many levels of indirection and generates. What do you think?

Yes, I also think having sane defaults seems to be the best approach and then selectively overwriting. Writing it by hand also SGTM. 

> 3. I will also add other command line options, such as `num_cores` or `num_clusters`, to the config. Do you agree that this would be nice to have? The order of precedence would then be:

Perfect!

One thing that came to my mind (I guess this is also somewhat important for `mempool`): 

- We somehow need a way to model the NUMA in our system, i.e., that a cluster/tile can access memory of its neighboring cluster with shorter latency than cluster/tile memory of tile further away, right?
- In `Manticore.json`/`occamy.json` above I think we also need to model the global (DRAM) memory.
- What could also work, is that we give Banshee a fully structural description of the system. Depending on the complexity we could also generate this with Python. We could also hard-code the structure of the system at this stage. I am fine with either approach.

Let me know what you think. Thanks again for your awesome work!

SamuelRiedel:
> That is really, really nice @SamuelRiedel!
> 
> > 1. [Manticore.json](https://github.com/SamuelRiedel/snitch/blob/71f1f0f281d21a014242a99d70ca8dbb9429005a/sw/banshee/manticore.json) is an example of how to configure Banshee. You can set the TCDM location and size, the control registers, and the latencies of the loads to different regions and each instruction. Is there anything else you would like to configure in the near future? Any suggestions on the format? For example, we could clearly split the memory latencies from the instruction latencies and instead specify the latency with the memory range.
> 
> Small nitpick. Can we re-name this `occamy.json`? Then we can further develop this together with the RTL in here.

Of course :+1:
> 
> > 1. How should we handle defaults? The current implementation requires a JSON configuration file, otherwise, the control register addresses are not set. This would break the previous flow because we have a new mandatory command line flag. We could either parse the `manticore.json` by default, but that prevents the binary from being relocatable. Or, we add a new struct in Rust that represents the default configuration, and we only parse the JSON to overwrite some values. I prefer the latter because it will not break anyone's flow and we have the default configuration as a 'documentation' of what can be set via JSON. However, this also means that we need to create a default configuration. The latency for each instruction can be [generated](https://github.com/SamuelRiedel/snitch/blob/71f1f0f281d21a014242a99d70ca8dbb9429005a/sw/banshee/src/riscv.rs#L2595) quite easily with `riscv-opcodes`. I suggest writing the architecture configuration by hand as a Rust struct to avoid too many levels of indirection and generates. What do you think?
> 
> Yes, I also think having sane defaults seems to be the best approach and then selectively overwriting. Writing it by hand also SGTM.

:+1:
> 
> > 1. I will also add other command line options, such as `num_cores` or `num_clusters`, to the config. Do you agree that this would be nice to have? The order of precedence would then be:
> 
> Perfect!
> 
> One thing that came to my mind (I guess this is also somewhat important for `mempool`):
> 
>     * We somehow need a way to model the NUMA in our system, i.e., that a cluster/tile can access memory of its neighboring cluster with shorter latency than cluster/tile memory of tile further away, right?

Absolutely. This will be a nightmare to tackle... On MemPool, we could calculate this based on the `hart_id` and the address. I was considering just writing a MemPool specific function that calculates the latency. But it would be way nicer to have this properly configurable. I don't really know how to tackle this tough. My best solution at the moment is to have an address range + latency for each core. But then we need to generate a huge configuration. If we can always calculate the latencies based on the hart_id and the memory_map, then we could also just implement a function per system in rust and configure which function to use.

> 
>     * In `Manticore.json`/`occamy.json` above I think we also need to model the global (DRAM) memory.
>

Yes. Currently, we just have `tcdm` which is allocated in the host memory directly, and "`dram`", which is just a hash-map capturing all other addresses. What you mean is an additional configuration describing the `dram` range and then generate some warnings if we make accesses outside any valid range?
> 
>     * What could also work, is that we give Banshee a fully structural description of the system. Depending on the complexity we could also generate this with Python. We could also hard-code the structure of the system at this stage. I am fine with either approach.
>

That would be quite nice, but I think that requires a lot of rework in Banshee to make sure that we could actually handle all the configurations.
> 
> 
> Let me know what you think. Thanks again for your awesome work!

Thanks for your feedback! :smile:

SamuelRiedel:
Ok I updated the way the configuration works and I rewrote the PR message. Please have a look. I suggest we tackle some open problems in a next PR. Namely adding a NUMA model.

sw/banshee/src/main.rs
```diff
@@ -93,6 +95,18 @@ fn main() -> Result<()> {
...
+                .long("configuration")
+                .takes_value(true)
+                .help("A configuration file describing the architecture"),
+        )
+        .arg(
+            Arg::with_name("create-configuration")
+                .long("create-configuration")
+                .takes_value(true)
+                .help("Write the default configuration to this file"),
+        )
```
zarubaf, line 109: Nice!

zarubaf (approved):
That looks very nice and solid to me! Thanks @SamuelRiedel

paulsc96 (approved):

PR #84, opened 2021-03-29
Banshee fix sdma strt

noah95:
the `dmcpy` instruction didn't pass the `cpu` argument to `banshee_dma_strt` which caused a compile error.

Further, a test is added that sensitizes this fix and also ensures correct, multiple 1D DMA transfers.

There is still another test `dma_simple` that is currently ignored. It can be removed since the new test mirrors the behavior but is written using `builtins` which are currently not supported in the upstream toolchain. What do you think, should we replace the `dma_simple` test or leave it be? It fails currently because of wrong `dm` instruction encoding.


paulsc96 (approved):
Looks good to me! as for the existing test, I think we can remove it provided the new one covers the same functionality.

Edit: I missed the part about the new toolchain. This is a bit more difficult. Keeping the existing test that does not work is of no use, but maybe we can find an interim solution to using the new test with the old toolchain, like preprocessor-guarded substitute definitions or something similar.

sw/banshee/src/tran.rs
```diff
@@ -2135,9 +2135,20 @@ impl<'a> InstructionTranslator<'a> {
...
+                    LLVMBuildBitCast(
+                        self.builder,
+                        self.section.state_ptr,
+                        LLVMPointerType(LLVMInt8Type(), 0),
+                        NONAME,
+                    ),
+                    rs1,
+                    rs2,
+                ],
+            ),
```
SamuelRiedel, line 2151: Do you know why the CPU pointer is an 8-bit pointer here? I think we could clean this up with the following changes just to make it a bit easier to understand.

```diff
diff --git a/sw/banshee/src/runtime/jit.ll b/sw/banshee/src/runtime/jit.ll
index 07bedce..d1f8903 100644
--- a/sw/banshee/src/runtime/jit.ll
+++ b/sw/banshee/src/runtime/jit.ll
@@ -34,7 +34,7 @@ declare void @banshee_ssr_eoi(%SsrState* %ssr)
 
 declare void @banshee_dma_src(%DmaState* writeonly %dma, i32 %lo, i32 %hi)
 declare void @banshee_dma_dst(%DmaState* writeonly %dma, i32 %lo, i32 %hi)
-declare i32 @banshee_dma_strt(%DmaState* %dma, i8* %cpu, i32 %size, i32 %flags)
+declare i32 @banshee_dma_strt(%DmaState* %dma, %Cpu* %cpu, i32 %size, i32 %flags)
 declare i32 @banshee_dma_stat(%DmaState* readonly %dma, i32 %addr)
 
 declare i32* @banshee_reg_ptr(%Cpu* %cpu, i32 %reg)
diff --git a/sw/banshee/src/tran.rs b/sw/banshee/src/tran.rs
index c825f41..a244e33 100644
--- a/sw/banshee/src/tran.rs
+++ b/sw/banshee/src/tran.rs
@@ -1348,17 +1348,7 @@ impl<'a> InstructionTranslator<'a> {
         let value = match data.op {
             riscv::OpcodeImm5RdRs1::Dmcpyi => self.section.emit_call(
                 "banshee_dma_strt",
-                [
-                    self.dma_ptr(),
-                    LLVMBuildBitCast(
-                        self.builder,
-                        self.section.state_ptr,
-                        LLVMPointerType(LLVMInt8Type(), 0),
-                        NONAME,
-                    ),
-                    rs1,
-                    imm,
-                ],
+                [self.dma_ptr(), self.section.state_ptr, rs1, imm],
             ),
             // _ => bail!("Unsupported opcode {}", data.op),
         };
@@ -2137,17 +2127,7 @@ impl<'a> InstructionTranslator<'a> {
             riscv::OpcodeRdRs1Rs2::Sra => LLVMBuildAShr(self.builder, rs1, rs2, name),
             riscv::OpcodeRdRs1Rs2::Dmcpy => self.section.emit_call(
                 "banshee_dma_strt",
-                [
-                    self.dma_ptr(),
-                    LLVMBuildBitCast(
-                        self.builder,
-                        self.section.state_ptr,
-                        LLVMPointerType(LLVMInt8Type(), 0),
-                        NONAME,
-                    ),
-                    rs1,
-                    rs2,
-                ],
+                [self.dma_ptr(), self.section.state_ptr, rs1, rs2],
             ),
             _ => bail!("Unsupported opcode {}", data.op),
         };
```

SamuelRiedel (changes requested):
Nice fix :). I don't fully understand the toolchain problem. With the current toolchain/repo, we could not generate the binary you added in this PR, is this correct? But the `dma_simple` test does not work, because they use the wrong instruction decoding, right? Could we fix the `dma_simple` test, by fixing the encoding?

sw/banshee/src/tran.rs
noah95: Nice catch, thanks! I just copy-pasted it and didn't notice that this could be simplified.

noah95:
> I don't fully understand the toolchain problem

The test I introduced uses the `llvm` toolchain I am working on, especially the `builtins` for SDMA. A `make` in `banshee/test` would fail with the wrong toolchain

> we could not generate the binary you added in this PR, is this correct?

correct

> But the `dma_simple` test does not work, because they use the wrong instruction decoding, right? 

yes

> Could we fix the `dma_simple` test, by fixing the encoding?

Yes that is possible :)

SamuelRiedel (approved):
Looks good to me now :+1:

PR #89, opened 2021-03-30
snitch_icache: Disable the prefetcher if there are multiple hits

SamuelRiedel:
If we have multiple early hits, the [instruction data](https://github.com/pulp-platform/snitch/blob/master/hw/ip/snitch_icache/src/snitch_icache_l0.sv#L166) will be garbage. Therefore, the prefetcher might calculate a garbage address and create a fetch to invalid memory. This PR disables the prefetcher when we have multiple early hits.

Alternatively, we could also modify the prediction logic to not try to predict the branch but just prefetch the next block. E.g. setting [`is_branch_taken` & `is_jal`](https://github.com/pulp-platform/snitch/blob/master/hw/ip/snitch_icache/src/snitch_icache_l0.sv#L290) to zero. This might be the better option,  but I would appreciate your opinion.

### DIFF
--- a/hw/ip/snitch_icache/src/snitch_icache_l0.sv
+++ b/hw/ip/snitch_icache/src/snitch_icache_l0.sv
@@ -273,7 +273,7 @@ module snitch_icache_l0 import snitch_icache_pkg::*; #(
   // pre-fetched the line yet and there is no other refill in progress.
   assign prefetcher_out.vld = enable_prefetching_i &
                               hit_any & ~hit_prefetch_any &
-                              ~pending_refill_q;
+                              hit_early_is_onehot & ~pending_refill_q;
 
   localparam int unsigned FetchPkts = CFG.LINE_WIDTH/32;
   logic [FetchPkts-1:0] is_branch_taken;


zarubaf:
Thanks @SamuelRiedel. I've just fixed the compile script for the `snitch_icache_l0_tb`. Going to `hw/ip/snitch_icache` and running `./util/compile.sh` and `./util/run_vsim.sh` should be enough to run the testbench on the l0 cache. If you could briefly, re-base and try if things still pass, that would be much appreciated :-)

Eventually, it would be nice to have these checks available through our internal GitlabCI, but I guess this is a story for another day :-)

Regarding the proposed fix, my first intuition would be to disable the prefetcher, since the next line predictor could make our prefetcher busy for the real prediction that we will make when we finally know the instructions, therefore increasing the overall latency. And in the best case, we would only be saving one cycle. Do you agree?

hw/ip/snitch_icache/src/snitch_icache_l0.sv
```diff
@@ -288,26 +288,28 @@ module snitch_icache_l0 import snitch_icache_pkg::*; #(
...
+            is_branch_taken[i] = ins_data[i*32+31];
+          end
+          riscv_instr::JAL: begin
+            is_jal[i] = 1'b1;
+          end
+          // we can't do anything about the JALR case as we don't
+          // know the destination.
+          default:;
+        endcase
+      end
```
zarubaf, line 312: Technically that would also work, not sure if it makes anything better though (except reducing diff size :-))
```suggestion
      // if we get a double hit produce an invalid RISC-V instruction
      unique casez (ins_data[i*32+:32] | hit_early_is_onehot)
        // static prediction
        riscv_instr::BEQ,
        riscv_instr::BNE,
        riscv_instr::BLT,
        riscv_instr::BGE,
        riscv_instr::BLTU,
        riscv_instr::BGEU: begin
          // look at the sign bit of the immediate field
          // backward branches (immediate negative) taken
          // forward branches not taken
          is_branch_taken[i] = ins_data[i*32+31];
        end
        riscv_instr::JAL: begin
          is_jal[i] = 1'b1;
        end
        // we can't do anything about the JALR case as we don't
        // know the destination.
        default:;
      endcase
```

SamuelRiedel: Not sure if this would work. We just set the LSB of the instruction to 1, which is anyways the case for the branches. But if we agree on just disabling the prefetcher altogether, this whole change is not necessary anyways.

zarubaf:
Great! We are in total agreement then :-) Did the L0 testbench pass for you?

SamuelRiedel:
I agree with just disabling the prefetch completely. Worst case, we just lose one cycle. I rebased the branch and tried out your testbench, which still passes.

zarubaf:
Awesome! Thanks a lot - merging once CI passes.

PR #90, opened 2021-04-01
ssr: Indirection RTL

paulsc96:
Reintegrates the SSR datapath RTL for streaming indirection as described [here](https://arxiv.org/abs/2011.08070). To verify this functionality, the standalone SSR testbench was extended with example streaming jobs sensitizing all configuration options.

We propagate the indirection parameters up to  `snitch_ssr`, but do _not yet_ propagate them further or add them to the cluster configurator. Since the new parameters have stable defaults, the elaborated hardware of existing systems should remain unchanged.

PRs for the following will be created separately:

* Software (kernels)
* Full parameterization and configurator integration
* Banshee support.


hw/ip/snitch_ssr/src/snitch_ssr_addr_gen.sv
```diff
@@ -55,23 +77,209 @@ module snitch_ssr_addr_gen import snitch_pkg::*; #(
...
+        idx_shift_sq  <= idx_shift_sd;
+        idx_base_sq   <= idx_base_sd;
+        idx_size_sq   <= idx_size_sd;
+        if (config_q.done) begin
+          idx_shift_q <= idx_shift_sd;
+          idx_base_q  <= idx_base_sd;
+          idx_size_q  <= idx_size_sd;
+        end
+      end
+    end
```
zarubaf, line 168: Would you mind switching that to the `` `FF`` macros? That would make it easier to change flip-flop implementation and is a general goal for the modules in here :-)

hw/ip/snitch_ssr/src/snitch_ssr_indirector.sv
```diff
@@ -0,0 +1,157 @@
...
+  // Index handshaking
+  assign idx_req_o.q_valid  = cfg_indir_i & idx_cred_left & ~natit_done_i;
+  assign natit_ready_o      = cfg_indir_i & idx_cred_left & idx_rsp_i.q_ready;
+
+  // Index FIFO: stores full unserialized words.
+  fifo_v3 #(
+    .FALL_THROUGH ( 1'b0          ),
+    .DATA_WIDTH   ( DataWidth     ),
+    .DEPTH        ( IndexCredits  )
+    ) i_idx_fifo (
```
zarubaf, line 98:
```suggestion
  ) i_idx_fifo (
```

zarubaf (approved):
That looks really nice - thanks a lot! Do you have an idea of the coverage we get with the testbench?

hw/ip/snitch_ssr/src/snitch_ssr_addr_gen.sv
paulsc96: I switched all `always_ff` blocks to `FF macros in 66f1d32343fb136e96516d2bae21445b4a98835d.

I also added more coverage options to the testbench start scripts. Looks OK overall except for a low toggle coverage (61%) in the indirector and address generator. This is probably because the simple / short test jobs don't toggle the datapath MSBs, but that can be improved if necessary.

zarubaf (approved):
Awesome! Very clean contribution, thanks!

PR #91, opened 2021-04-02
occamy: Add Xilinx wrapper

niwis:
Vivado's IP Integrator requires flattened AXI ports. Add a wrapper for Occamy that generically transforms the AXI ports to the required format.


hw/system/occamy/src/occamy_xilinx.sv
```diff
@@ -0,0 +1,74 @@
...
+  input  logic             i2c_sda_i,
+  output logic             i2c_sda_en_o,
+  output logic             i2c_scl_o,
+  input  logic             i2c_scl_i,
+  output logic             i2c_scl_en_o,
+
+
+  /// PCIe Ports
+  `AXI_FLAT_MASTER_PORT(pcie, logic [47:0], logic [4:0], logic [511:0], logic [63:0], logic [0:0])
+  ,`AXI_FLAT_SLAVE_PORT(pcie, logic [47:0], logic [4:0], logic [511:0], logic [63:0], logic [0:0])
```
niwis, line 53: If the first comma is placed at the end of the previous line, variable fails.

zarubaf: Would it work if we use a "port" template here that we place in `util/solder/solder.axi_flatten_port.sv.tpl`? `emit_flat_[master|slave]_port("pcie")` could then just emit the template (similar to the CDC logic for example):
```python
tpl = templates.get_template("solder.axi_flatten_port.sv.tpl")
        context.write(tpl.render_unicode(...))
```

We could also handle the direction of the ports programmatically in that template (i.e., swapping directions) and reducing code duplication in the `axi_flat.sv` :-)

util/occamygen.py
```diff
@@ -255,6 +257,15 @@ def main():
         code = re_trailws.sub("", code)
         file.write(code)
 
+    with open(args.XILINX_SV, "w") as file:
+        code = tpl_xilinx.render_unicode(
+            module=solder.code_module['default'].replace("\n", "\n  "),
```
zarubaf, line 262:
```suggestion
```

I don't think you need the `module` context here.

zarubaf (commented):
Thanks a lot, that is great! Small nitpick about the way we generate the ports, but looking good already! :-)

niwis:
I moved the flat AXI port declaration from the macro to a template :-)

hw/system/occamy/src/axi_flat.sv
```diff
@@ -0,0 +1,117 @@
...
+  assign rsp.b.id     = m_axi_``pat``_bid;       \
+  assign rsp.b.resp   = m_axi_``pat``_bresp;     \
+  assign rsp.b.user   = m_axi_``pat``_buser;     \
+                                                 \
+  assign rsp.r_valid  = m_axi_``pat``_rvalid;    \
+  assign rsp.r.id     = m_axi_``pat``_rid;       \
+  assign rsp.r.data   = m_axi_``pat``_rdata;     \
+  assign rsp.r.resp   = m_axi_``pat``_rresp;     \
+  assign rsp.r.last   = m_axi_``pat``_rlast;     \
+  assign rsp.r.user   = m_axi_``pat``_ruser
```
zarubaf, line 62: Super small nitpick, afaik we usually include a trailing semicolon. I think that is the standard in UVM and worst case if it is called with another semicolon it is still valid SV, while in the other way, it can lead to "subtle" bugs.
```suggestion
  assign rsp.r.user   = m_axi_``pat``_ruser;
```

hw/system/occamy/src/axi_flat.sv
```diff
@@ -0,0 +1,117 @@
...
+  assign s_axi_``pat``_bid     = rsp.b.id;       \
+  assign s_axi_``pat``_bresp   = rsp.b.resp;     \
+  assign s_axi_``pat``_buser   = rsp.b.user;     \
+                                                 \
+  assign s_axi_``pat``_rvalid  = rsp.r_valid;    \
+  assign s_axi_``pat``_rid     = rsp.r.id;       \
+  assign s_axi_``pat``_rdata   = rsp.r.data;     \
+  assign s_axi_``pat``_rresp   = rsp.r.resp;     \
+  assign s_axi_``pat``_rlast   = rsp.r.last;     \
+  assign s_axi_``pat``_ruser   = rsp.r.user
```
zarubaf, line 117:
```suggestion
  assign s_axi_``pat``_ruser   = rsp.r.user;
```

hw/system/occamy/src/occamy_xilinx.sv.tpl
```diff
@@ -0,0 +1,76 @@
...
+  output logic             i2c_sda_en_o,
+  output logic             i2c_scl_o,
+  input  logic             i2c_scl_i,
+  output logic             i2c_scl_en_o,
+
+
+  /// PCIe Ports
+  <% soc_wide_xbar.out_pcie \
+        .emit_flat_master_port(context, "pcie", ",\n\n") \
+        .emit_flat_slave_port(context, "pcie", "\n")
```
zarubaf, line 54: That should be, right?
```suggestion
  <% soc_wide_xbar.out_pcie \
        .emit_flat_master_port(context, "pcie", ",\n\n") \
        .emit_flat_slave_port(context, "pcie", "\n")
```
```suggestion
  <% soc_wide_xbar.out_pcie \
        .emit_flat_master_port(context, "pcie", ",\n\n")
        soc_wide_xbar.in_pcie.emit_flat_slave_port(context, "pcie", "\n")
```

hw/system/occamy/src/occamy_xilinx.sv.tpl
```diff
@@ -0,0 +1,76 @@
...
+  // AXI ports of Occamy top-level
+  ${soc_wide_xbar.out_pcie.req_type()} pcie_axi_req_o;
+  ${soc_wide_xbar.out_pcie.rsp_type()} pcie_axi_rsp_i;
+
+  ${soc_wide_xbar.out_pcie.req_type()} pcie_axi_req_i;
+  ${soc_wide_xbar.out_pcie.rsp_type()} pcie_axi_rsp_o;
+
+  // Assign structs to flattened ports
+  `AXI_FLATTEN_MASTER(pcie, pcie_axi_req_o, pcie_axi_rsp_i);
+  `AXI_FLATTEN_SLAVE(pcie, pcie_axi_req_i, pcie_axi_rsp_o);
```
zarubaf, line 69:
```suggestion
  `AXI_FLATTEN_MASTER(pcie, pcie_axi_req_o, pcie_axi_rsp_i)
  `AXI_FLATTEN_SLAVE(pcie, pcie_axi_req_i, pcie_axi_rsp_o)
```

zarubaf (commented):
Really like it! Especially the explicit nature :-) This will really pay-off once we add the HBI ports :-)

hw/system/occamy/src/occamy_xilinx.sv.tpl
zarubaf: I just saw that I got this wrong on the top-level as well :-) The "in" port should probably have a smaller ID width.

niwis: Sure! I updated both the wrapper and the top-level.

zarubaf (approved):
Awesome, thanks @niwis!

PR #97, opened 2021-04-06
Updated snrt.h so that it compiles with c++

diorga:
This small fix is needed if you want to compile with C++, instead of C.


zarubaf (approved):
That's great! Thanks! Would you mind re-running `clang-format` to make CI happy :-)

diorga:
Done!

PR #98, opened 2021-04-07
banshee: Reset FREP repetition counter on each invocation

paulsc96:
When changing the FREP implementation to using a stacked repetition counter, I forgot to ensure the counter is reset to 0 on every invocation. This explained the faulty behavior of repeated FREPs @noah95 observed.

I added a simple sensitizing test. @noah95: could you rebuild the existing `ssr-multi-use` with the second FREP commented in to check whether it works now, and commit the resulting binary on this branch? I have no (known-good) LLVM toolchain :)


noah95 (approved):
Nice, thanks for the fix!

SamuelRiedel (approved):
Looks good from my side too. Thanks.

PR #100, opened 2021-04-13
cluster-cfg: Consistent endianness in config, script, and template

paulsc96:
The generated cluster wrappers currently have inconsistent endianness for packed (flattened) and unpacked parameter arrays. 

Additionally, the existing cluster config files suggest little-endian hive and core lists, but the parsing scripts iterate through them in regular big-endian fashion.

This PR ensures consistency and correct wrappers by changing endianness in the existing configs to big-endian and reversing flattened parameter arrays.


zarubaf:
Thanks a lot. I think that way it is much more intuitive and - most importantly - consistent!

zarubaf (approved):
LGTM

PR #101, opened 2021-04-13
ssr: Parameterize fully and add to cluster configuration

paulsc96:
This PR exposes almost all previously hardcoded or locally set SSR parameters through a configuration struct and propagates them to the cluster top-level. It also extends the existing cluster configuration schema, template, and script to set these new configurations.

Additionally, the SSR testbench was extended and two small parameterization bugs were fixed:
* The `snitch_ssr_indirector` assumed an inverted credit count, meaning it worked only incidentally with `IndexCredits=2`.
* Some SSR outputs in `snitch_cc` were not tied off correctly when `Xssr` is set to `0`.

@zarubaf Could you add a commit rebuilding the schema doc?


zarubaf (approved):
Great, thanks a lot @paulsc96 - I don't have much to add here, it looks very good!

paulsc96:
Thanks for the rebuild :)

PR #102, opened 2021-04-14
util/Makefrag: Catch failing dependency resolution

niwis:
Catch the `no matches` error of `ls` when including the Makefrag in a directory without any trace files.


zarubaf (approved):
You are my hero, that was annoying me all the time but I never thought about fixing it :-D

PR #106, opened 2021-04-15
banshee: Fake uart to stderr with formatting and hartid

noah95:
Allows to `--trace` at the same time as observing the output of the fake uart. Additionally, all printed lines are prepended with the source-hard-id

image


paulsc96 (approved):
This is a nice and light alternative to handling concurrent file IO in banshee :)

SamuelRiedel (approved):
LGTM. Thanks

Issue #105, opened 2021-04-15
banshee non-zero TCDM offset fails initialization and loads


noah95:
When the `tcdm.start` parameter in the banshee configuration is set to a non-zero value, the ELF data preloading and the fast loads do not access the same region.

The tcdm is initialized here

https://github.com/pulp-platform/snitch/blob/457b30533107ebfa8476a54aa07643450b01539d/sw/banshee/src/engine.rs#L321

where the address is absolute.

The fast access to the tcdm 
https://github.com/pulp-platform/snitch/blob/457b30533107ebfa8476a54aa07643450b01539d/sw/banshee/src/tran.rs#L2815
however, is relative to the tcdm start address.

Either, the `LLVMBuildSub` in `emit_tcdm_check` has to be removed, or the offset be subtracted when the tcdm is initialized.

PR #107, opened 2021-04-15
banshee: Fix tcdm initialization by offsetting access

noah95:
As described in #105, if a non-zero TCDM base is used, the memory initialization breaks


SamuelRiedel (approved):
Perfect. Thanks a lot for fixing this.

PR #109, opened 2021-04-16
banshee: Enable TCDM access from `banshee_load` and `banshee_store` calls

paulsc96:
This enables a consistent memory view with the TCDM mapped to its specified location for the core fallback and DMA access cases. Therefore, this also enables DMA copies to and from the TCDM and between TCDM locations.

As of now, the implementation is _not_ safe for multiple writers, as this would require significant changes in the way TCDM accesses are handled. However, this is likely not an issue as cores access the TCDM directly and external  / DMA accesses are not expected to provide safe concurrent writes.

The existing `dma_simple` test was adapted to demonstrate a minimal intra-TCDM DMA transfer.


SamuelRiedel (approved):
👍  LGTM thanks.

noah95:
Thanks for the effort @paulsc96 but that doesn't solve it for me somehow. This program still fails with non-zero tcdm in banshee_config
```c
static uint32_t a[] = {1,2,3,4,5,6,7,8}; // 00300000 <main.a>:
static uint32_t b[] = {0,0,0,0,0,0,0,0}; // 003000c8 <main.b>

__builtin_sdma_start_oned((uint64_t)a, (uint64_t)b, 8*sizeof(uint32_t), 0);
__builtin_sdma_wait_for_idle();

// tcdm:
//     start: 0x00300000
for (int i = 0; i < 8; ++i) printf("a %d b %d\n",a[i],b[i]);
 // hart-000  a 1 b 0
 // hart-000  a 2 b 0
 // hart-000  a 3 b 0
 // hart-000  a 4 b 0

// tcdm:
//     start: 0x00000000
for (int i = 0; i < 8; ++i) printf("a %d b %d\n",a[i],b[i]);
 // hart-000  a 1 b 1
 // hart-000  a 2 b 2
 // hart-000  a 3 b 3
 // hart-000  a 4 b 4
```

sw/banshee/src/engine.rs
```diff
@@ -498,6 +498,16 @@ impl<'a, 'b> Cpu<'a, 'b> {
             } // cluster_base_hartid
             x if x == self.engine.config.address.cluster_num => self.engine.num_clusters as u32, // cluster_num
             x if x == self.engine.config.address.cluster_id => self.cluster_id as u32, // cluster_id
+            // TCDM
+            x if x >= self.engine.config.memory.tcdm.start
+                && x < self.engine.config.memory.tcdm.end =>
+            {
+                let word_addr = addr / 4;
+                let word_offs = addr - 4 * word_addr;
```
noah95, line 506:
```suggestion
                let word_addr = (addr - self.engine.config.memory.tcdm.start) / 4;
                let word_offs = (addr - self.engine.config.memory.tcdm.start) - 4 * word_addr;
```

sw/banshee/src/engine.rs
```diff
@@ -554,6 +564,23 @@ impl<'a, 'b> Cpu<'a, 'b> {
...
                 }
             }
+            // TCDM
+            // TODO: this is *not* thread-safe and *will* lead to undefined behavior on simultaneous access
+            // by 2 harts. However, changing `tcdm_ptr` to a locked structure would require pervasive redesign.
+            x if x >= self.engine.config.memory.tcdm.start
+                && x < self.engine.config.memory.tcdm.end =>
+            {
+                let word_addr = addr / 4;
+                let word_offs = addr - 4 * word_addr;
```
noah95, line 574:
```suggestion
                let word_addr = (addr - self.engine.config.memory.tcdm.start) / 4;
                let word_offs = (addr - self.engine.config.memory.tcdm.start) - 4 * word_addr;
```

sw/banshee/src/engine.rs
noah95: we need to subtract the TCDM start offset here

sw/banshee/src/engine.rs
noah95: we need to subtract the TCDM start offset here

noah95 (changes requested):

noah95:
e6e3923  fixes it, thanks. ready to merge :+1:

noah95 (approved):

PR #111, opened 2021-04-22
snitch_cluster: Re-add `wake-up` register

zarubaf:
Re-adds the wake-up register. Following the Mempool implementation. As I don't have any meaningful set-up to test the implementation I'd appreciate your feedback @noah95


noah95:
Thanks so much for this! All my code with omp runtime works with this implementation of the `wake-up` register.

The address for the barrier register changed as well correct? Then there should be this change in the `snRuntime`:

- [ ] Change `0x28` to `0x30`

https://github.com/pulp-platform/snitch/blob/e878d4dee0343443421e943e1f7632e96e8c55f2/sw/snRuntime/src/start_cluster.c#L76

zarubaf:
Thanks for being so attentive, yes that should be shifted as well. I somehow thought that it is derived from the generated C header file, apparently I was wrong.

noah95:
Nice, the rest looks good to me :+1:

PR #112, opened 2021-04-22
banshee: ISSR-FREP sparse-dense linear algebra tests

paulsc96:
This tests adds SpMV and SpMM tests alongside the existing SpVV test to validate ISSRs used in combination with FREP in Banshee.

These new tests use the `snitch-llvm` toolchain, for which a simple build setup is added. The existing `spvv-issr-frep` and `dotp-ssr-frep` tests are ported to this chain and their SSR configuration mechanism is updated to `Xssr` configuration instructions instead of the legacy memory-mapped configuration interface.


SamuelRiedel (approved):
LGTM. :+1:

PR #113, opened 2021-04-24
occamy: Add APB port to HBM controller

niwis:
This PR adds an APB port to Occamy that can be connected to the HBM's controller. Within Occamy, the APB port is connected to the peripheral xbar.
To keep the APB aligned with the rest of the system, `solder` is extended by APB primitives. Furthermore, the APB repository of `pulp-platform` is included and used. On the `occamy_xilinx` level, the APB port is exploded, so that it is recognised by Vivado.


hw/vendor/pulp_platform_apb.vendor.hjson
```diff
@@ -0,0 +1,17 @@
...
+
+  upstream: {
+    url: "https://github.com/pulp-platform/apb.git",
+    rev: "master",
+  },
+
+  patch_dir: "patches/pulp_platform_apb",
+
+  exclude_from_upstream: [
+  ]
```
niwis, line 16: Perhaps we can/should exclude some of the files?

hw/system/occamy/src/reg_to_apb.sv
```diff
@@ -0,0 +1,42 @@
...
+  assign apb_req_o.psel    = reg_req_i.valid;
+  assign apb_req_o.penable = reg_req_i.valid;
+  assign apb_req_o.pstrb   = reg_req_i.wstrb;
+  assign apb_req_o.pprot   = 3'b010;            // 0: unprivileged, 1: non-secure, 0: data
+
+  assign reg_rsp_o.ready   = apb_rsp_i.pready;
+  assign reg_rsp_o.error   = apb_rsp_i.pslverr;
+  assign reg_rsp_o.rdata   = apb_rsp_i.prdata;
+
+endmodule
```
niwis, line 42: I guess this file should eventually be moved to https://github.com/pulp-platform/register_interface

hw/system/occamy/src/reg_to_apb.sv
```diff
@@ -0,0 +1,42 @@
...
+  // APB interface
+  output apb_req_t apb_req_o,
+  input  apb_rsp_t apb_rsp_i
+);
+
+  assign apb_req_o.paddr   = reg_req_i.addr;
+  assign apb_req_o.pwrite  = reg_req_i.write;
+  assign apb_req_o.pwdata  = reg_req_i.wdata;
+  assign apb_req_o.psel    = reg_req_i.valid;
+  assign apb_req_o.penable = reg_req_i.valid;
```
zarubaf, line 34: Are you sure that the APB protocol is implemented correctly here? APB has two phases (https://web.eecs.umich.edu/~prabal/teaching/eecs373-f12/readings/ARM_AMBA3_APB.pdf 2.2 - 2.4): 

1. Setup phase (1 cycle): `psel`, `paddr`, `pwrite`, `pwdata` become high and stable.
2. Access phase (2 cycle): `penable` goes high. In the same cycle or in any of the following cycles `pready` goes high. Which terminates the transaction.

I think we'll need a flop to distinguish between the phases at least and gate the required signals (`psel`, `penable`, and `reg_rsp_o.ready`).

Does that make sense?

hw/system/occamy/src/reg_to_apb.sv
zarubaf: Yes, please! Alternatively, we could also push it into the `apb` repo. Not sure which direction makes sense. We can decide later together with @accuminium.

hw/vendor/pulp_platform_apb.vendor.hjson
zarubaf: Maybe the `scripts` folder?

zarubaf (changes requested):
Super clean contribution @niwis, thanks a lot. Only some doubts regarding the APB protocol implementation - please double-check that with the specification. 

The rest is perfect!

zarubaf:
Right now I was wondering whether we could actually attach the APB to an AXI crossbar we instantiate in the block design (driven by the PCIe master port). We are going to need this crossbar anyway (for the SPI and UART) and I am pretty sure Xilinx has the AXI to APB protocol converter.

Rationale: Right now I can't see that our ASIC HBM controller is going to have a similar port so I don't think we'll need that port on the shared RTL.

Does that make sense?

hw/system/occamy/src/reg_to_apb.sv
niwis: Good catch, I meant to check that.

niwis:
Sure, if the ASIC doesn't use it, this is probably not the right place for it. Do you think we can use parts of this PR anyways (e.g. the APB support)? In this case, I would open a new PR.

zarubaf:
Most definitely! I am pretty sure we are going to need the APB support sooner than later. So that is useful and we should merge it.

zarubaf:
@niwis could you give this a look again? In particular, I am asking because we likely are going to need the APB for the HBI portion that to configure the PHY and controller, right @stmach.

I think the only "missing" point was the two-cycle handling of APB. Would that work for you @niwis?

niwis:
Sure! I will try this week.

niwis:
Closed in favour of https://github.com/pulp-platform/snitch/pull/136.

PR #114, opened 2021-04-27
Fix/wide narrow assertion

zarubaf:
The comment in the file was correct in specifying that when the wide port is
selected there must always be an immediate grant (i.e., the wide mux has
priority over the SRAM bank and there can't be any stalling mechanism).

The assertion however, checked the property on all requests. Atomics (which
can only occur on the narrow port) violate this property since they require
a few cycles of RMW behavior.


noah95 (approved):
Makes sense :+1: this fixes the assertions in my use of the atomics, thanks!

PR #116, opened 2021-04-28
testbench: Increase DMA memory transactions to omit DMA deadlock

noah95:
Transaction issued by the DMA larger than 4K might deadlock. Increasing the transaction buffer fixes this in my case.

@zarubaf @accuminium do you think we can leave it like that for now?


zarubaf:
Yes absolutely! I’ll open an issue about the fundamental problem later and will take care of it :-) Thanks 🙏

zarubaf (approved):

PR #118, opened 2021-04-28
occamy: Generate a preliminary device tree

zarubaf:
This commit adds the ability to generate a device tree from a given
solder description. That should help to keep the DTS aligned to the
address map.

The feature set is quite rough and preliminary but might almost be
sufficient to boot a bare-bones Linux.


niwis (approved):
Looks good!

PR #120, opened 2021-04-29
occamy: Prevent duplicate address map entries for HBM

niwis:
Do not create two address map entries for the first two HBM channels (unless this was intended).


zarubaf:
That was indeed indented, sorry for the bad comment situation. Rationale was that we can use them as an alias to also boot with the 32 bit cores without setting up address translation - does that make sense?

niwis:
Yes, that makes sense!

PR #129, opened 2021-05-11
banshee: Fix wrong ssr repeat assignment

noah95:
The ssr configuration instruction was writing the repeat value to the `repeat_count` field instead of `repeat_bound`. This enables now to use the SSR repeat feature


paulsc96 (approved):
Very nice catch! LGTM

PR #130, opened 2021-05-12
occamy: Linux bringup on FPGA

zarubaf:
A couple of first steps for Linux FPGA boot. Still not successful, but I've figured that those changes are pretty uncontroversial and they can already go :-)


niwis (approved):
Looks good to me! :-)

PR #132, opened 2021-05-17
snitch: Decode `fdiv` and `fsqrt` as illegal

zarubaf:
As we do not support those units as of now it's less error prone to
decode them as illegal. In that way people at least know that they are
not implemented and SW can handle the missing units more gracefully.


zarubaf:
CC @noah95

PR #134, opened 2021-05-18
Revise CLINT implmentenation

zarubaf:
- Changes the CLINT to use lowRISC's reggen tool.
- Update UART device tree settings.


niwis (approved):
Perfect! Could we actually also integrate this CLINT in Ariane? Probably we would also need a `regbus`...

zarubaf:
Yes, I think that would make sense. For the initial bring-up, I've moved it from the IP's directory mainly because we would have ended up with a lot of diff :-) The CVA6 repo is also re-organizing itself at the moment but I guess we should move it to the dedicated `clint` repo that we have :-)

paulsc96:
> I guess we should move it to the dedicated `clint` repo that we have :-)

I was wondering what to do with this convenient regbus CLINT. Well, don't mind if I do just that :+1:

PR #137, opened 2021-05-20
occamy: Add scratchpad memory fixes #135

### LINKED ISSUE #135: occamy: Add 256 kByte of SRAM
For proper bootloading, we need to add a bit of extra SRAM. The ports should already be provisioned.

zarubaf:
For booting and fall-back purposes we want to include a bit of on-chip SRAM.


niwis (approved):
Makes sense, looks good to me!

PR #139, opened 2021-05-22
occamy/fpga: Make ILA excludable

niwis:
The current ILA is pretty large, also because it probes an AXI bus with a 512 bit data width. Often, the ILA is not needed or usable, e.g. when the JTAG port is occupied by the debugger. Excluding the ILA accelerates compile time by ~1h and reduces BRAM usage by ~40%.

This excludes the ILA per default. It can be included using the `make` variable `DEBUG` (e.g. `make fpga DEBUG=1`).


zarubaf (approved):
That's great!

PR #145, opened 2021-05-26
snitch_cluster: Add optional fp div and sqrt unit

zarubaf:
By setting the parameter in the cluster config file you should be able to get `div` and `sqrt` features for all available fp formats. I've just gave this a test with a very crude assembly smoke test so your feedback is much appreciated.

Also, note that the current FP divsqrt unit is known to be wrong about rounding at some occasions :-)


noah95 (approved):

noah95:
Oh, but i never tested `fsqrt.d`

zarubaf:
> Oh, but i never tested `fsqrt.d`

Probably enough for the moment.

> Is there a list of "proven" configurations where we should mention that this is untested?

That is a good point, like a warning or a notion that this is untested and unstable because of known divider issues. I'll try to come up with something. Thanks for the input.

PR #149, opened 2021-06-03
Fix atomic memory operations on global memory

zarubaf:
The specification was outdated with some of the implementations. I've
opted to signal all atomics with `write` set to `0` now as this seemed
to be the most straightforward way of implementing it.

Additionally, I've adapted the specification and added assertions to
capture this.

CC @fischeti


zarubaf:
@fischeti you confirmed that this solved your issue and you can perform atomics now, right? So that should be good to merge I guess?

fischeti:
Yes, I have tested atomic operations and this fix has resolved my previous issues.

PR #154, opened 2021-06-15
banshee: Add Load-reserved/Store-conditional

SamuelRiedel:
Implement RISCV's load-reserved and store-conditional by adding an extra register to hold the `compare` value. The LR will be replaced with a load storing an additional copy of the value in the `compare` register. The SC is replaced by a CAS using the `compare` register as its third operand. This is a very efficient implementation. However, it allows for potential ABA problems.


zarubaf:
Ping @paulsc96 - shall we merge?

paulsc96 (approved):

PR #155, opened 2021-06-16
ssr: DC compatibility fixes

zarubaf:
I was running the code through DC and apparently, it has problems with the following lines. I remember @paulsc96 telling me something similar the other day. Can you have a look at whether this is aligned with your expectations?

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -6,6 +6,9 @@
 # The DMA is owned by @thommythomaso
 hw/ip/snitch_dma @thommythomaso
 
+# SSRs are owned by @paulsc96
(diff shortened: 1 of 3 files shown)


paulsc96:
Yes, this is exactly what I fixed. Sorry about the delay on this. The pragma in the `FFLARNC` macro requires a proper signal and not an expression, so this addresses the issue.

paulsc96 (approved):

paulsc96:
This is fine, arguably even better. I didn't assign the LHS to signals to minimize the diff, but for consistency, all macro parameters should be signals and not expressions, also to prevent future issues with the macros.

PR #157, opened 2021-06-16
container: Add LLVM toolchain to the Snitch Docker container

noah95:
The `build-docker` run can be triggered manually and the LLVM toolchain version manually specified. The default value is `latest` which will be translated to the most recent tag in [snitch-llvm](https://github.com/pulp-platform/snitch-llvm/releases/tag/12.0.0-snitch-0.1.0-rc1)

### DIFF
--- a/.github/workflows/build-docker.yml
+++ b/.github/workflows/build-docker.yml
@@ -29,3 +29,5 @@ jobs:
         file: util/container/Dockerfile
         push: true
         tags: ghcr.io/pulp-platform/snitch:latest
+        build-args: |
+          SNITCH_LLVM_VERSION=latest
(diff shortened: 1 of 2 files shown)


noah95:
Was hoping we could test the `build-docker` action before merging if I pushed to this repo instead of my fork but i can't trigger the action. Can you @zarubaf ?

zarubaf:
Hm, I think that action is triggered only on master (to avoid overriding the published container). Let me investigate whether there is any possibility to trigger that by hand.

zarubaf:
Okay, I've added the option to trigger the `build-docker` workflow manually in #159. I think that also means you need to re-base your PR and let's see whether that solves the issue at hand. As you have write access on the repo, you should be able to trigger the workflow yourself (see https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow#configuring-a-workflow-to-run-manually).

noah95:
Ok this works now. The latest Snitch image can compile and run my functional regression tests as can be seen in [this run](https://github.com/pulp-platform/snitch-toolchain-cd/runs/2845960037?check_suite_focus=true). Ready to merge

zarubaf:
Great, thanks!

PR #158, opened 2021-06-16
Add `known` assertion on regfile write port

zarubaf:
X's have hunted us a couple of times already - this PR add assertions to check that we are not writing-back any X's on the floating or integer register file. That should at least track where we are introducing unknown values.


fischeti:
Thanks! The assertions are working.

PR #160, opened 2021-06-17
ssr: Fix missing scoreboard and wrong timin in config reads

noah95:
Two bugs in the RTL prevented the use of the `scfgr`/`scfgri` instructions

- A missing scoreboard connection to keep track of the destination register value
- Timing in the `stream_to_mem` converter. The SSR respond with the read value in the same cycle as the request is made. Previously, the `id` and `valid` signals were delayed a cycle but not the `data`. This change removes the delay such that the response comes in the same cycle as the request

This now allows SSR barriers such as 
```c
  asm volatile(
    "1: \n"
    "scfgri t0, 0\n"        // reat status word
    "srli   t0, t0, 31\n"   // extract done bit 31
    "beqz   t0, 1b\n"       // repeat if done bit not set
    ::: "t0");
```

Is the timing ok like this or should we insert a cut on the response?

### DIFF
--- a/hw/ip/snitch/src/snitch.sv
+++ b/hw/ip/snitch/src/snitch.sv
@@ -1936,6 +1936,7 @@ module snitch import snitch_pkg::*; import riscv_instr::*; #(
         if (Xssr) begin
           acc_qreq_o.addr = SSR_CFG;
           acc_qvalid_o = valid_instr;
+          acc_register_rd = 1'b1;
         end else illegal_inst = 1'b1;
       end
       SCFGWI: begin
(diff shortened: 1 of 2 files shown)


zarubaf:
> Two bugs in the RTL prevented the use of the `scfgr`/`scfgri` instructions
> 
> * A missing scoreboard connection to keep track of the destination register value

Perfect.

> * Timing in the `stream_to_mem` converter. The SSR respond with the read value in the same cycle as the request is made. Previously, the `id` and `valid` signals were delayed a cycle but not the `data`. This change removes the delay such that the response comes in the same cycle as the request

Interesting that this works, I would have thought that the `stream_to_mem` requires one cycle latency. Another fix would be to also delay `data`, right? I think I'd me more comfortable in getting such a fix in - not so much regarding the timing, but at least in the upstream `reqrsp` protocol a cycle delay between a request and a response is necessary. Does that make sense?

noah95:
Sure, this has now 1 cycle latency

image

zarubaf (approved):
Thanks, LGTM. The latency shouldn't be a big issue I guess.

PR #161, opened 2021-06-17
Remove atomics from DMA path

zarubaf:
The DMA has the potential to lock-up crossbars because of too many outstanding transactions. This is due the `aw`/`ar` dependency created by the support for atomics. This PR makes the support optional and deactivates it for the DMA network.


niwis (commented):
Looks good to me :-)

Issue #152, opened 2021-06-14
Read-modify-write (RMW) module


zarubaf:
In order to be able to protect an SRAM using ECC, only full-word accesses can be made as the parity word is assembled over the entire word. It is still beneficial to allow for byte-wise accesses though.

The idea is to create a module that contains the following interface on the in-side (minus atop):  https://github.com/pulp-platform/snitch/blob/d59892280a1ee5036ad174e600cd00ade36f79bb/hw/system/occamy/src/occamy_top.sv#L1637-L1645

If the module detects a write access without the full write mask set a state machine should do a read access, followed by a partial merge (dependent on the write mask) and a consecutive write. The output interface of the module looks similar to the input interface minus the write-mask (which is implicitly set to `'1`).

PR #164, opened 2021-06-21
Add: Read-modify-write (RMW) module

fischeti:
Read-modify-write adapter for byte-wise read and write accesses which addresses #152 . For byte-wise read accesses, the adapter just returns the masked data. For byte-wise write accesses, the adapter stalls the original request while performing a full-width read access followed by a masked modify and full-width write

The input and of the module consist of the following signals (the output signals are mirrored)
```sv 
   // Request-side channel
   input logic  mem_req_i,
   output logic mem_gnt_o,
   input        addr_t mem_addr_i,
   input        data_t mem_wdata_i,
   input        strb_t mem_strb_i,
   input logic  mem_we_i,
   output logic mem_rvalid_o,
   output       data_t mem_rdata_o
```

The protocol currently takes the following assumptions:
1. A valid handshake occurs when `req` and `gnt` are high. 
2. The data is read/written one cycle after a valid handshake and the `rvalid` signal is high for both read/write requests
3. A `gnt` is raised at the earliest one cycle after the `req` is raised

Open questions:
- Are the protocol assumptions correct? 
- Naming and location etc.: the module currently has its own folder `ip/spm_interface`. Could probably be placed somewhere else and maybe named differently. The interface is currently only used for testing purposes
- Which license header to use?

### DIFF
--- a/hw/ip/spm_interface/src/spm_rmw_adapter.sv
+++ b/hw/ip/spm_interface/src/spm_rmw_adapter.sv
@@ -50,22 +50,22 @@ module spm_rmw_adapter
   mem_data_t masked_wdata_q, masked_wdata_d, masked_data;
   assign masked_data = (mem_wdata_i & mask_q) | (mem_rdata_i & ~mask_q);
 
-  logic         partial_write;
+  logic partial_write;
   assign partial_write = mem_valid_i & mem_we_i & ~(&mem_strb_i);
 
   state_e req_state_q, req_state_d;
 
   cnt_t cnt_q, cnt_d;
-  logic         rmw_ready, txns_ready;
+  logic rmw_ready, txns_ready;
   assign rmw_ready = (cnt_q == '0);
   assign txns_ready = cnt_q < MaxTxns;
 
   always_comb begin
     cnt_d = cnt_q;
-    if (mem_valid_o && mem_ready_i) begin
+    if (mem_valid_i && mem_ready_o) begin
       cnt_d++;
     end
-     if (mem_rvalid_i) begin
+     if (mem_rvalid_o) begin
        cnt_d--;
      end
   end
@@ -75,38 +75,7 @@ module spm_rmw_adapter
       mask_d[i] = mem_strb_i[i/8];
     end
 
-    masked_wdata_d = (mem_rvalid_i)? masked_data : masked_wdata_q;
-  end
-
-  always_comb begin
-
-    req_state_d = req_state_q;
-
-    unique case (req_state_q)
-
-      NORMAL: begin
-        // If partial memory access is detected perform RMW_READ
-        if (partial_write && mem_ready_i && rmw_ready) begin
-          req_state_d = RMW_READ;
-        end
-      end
-
-      RMW_READ: begin
-        // Wait for full access read request to be granted
-        if (mem_rvalid_i) begin
-          req_state_d = RMW_MODIFY_WRITE;
-        end
-      end // case: RMW_READ
-
-      RMW_MODIFY_WRITE: begin
-        if (mem_rvalid_i) begin
-          req_state_d = NORMAL;
-        end
-      end
-
-      default: ;
-
-    endcase
+    masked_wdata_d = masked_data;
   end
 
   always_comb begin
@@ -123,12 +92,16 @@ module spm_rmw_adapter
     mem_rvalid_o = mem_rvalid_i;
     mem_rdata_o = mem_rdata_i;
 
+    req_state_d = req_state_q;
+
+
     unique case (req_state_q)
 
       NORMAL: begin
 
         // If access is bitwise, generate RMW_READ request
         if (partial_write && mem_ready_i && rmw_ready) begin
+          req_state_d = RMW_READ;
           mem_we_o = '0;
         end
       end // case: NORMAL
@@ -141,6 +114,7 @@ module spm_rmw_adapter
         mem_ready_o = 1'b0;
 
         if (mem_rvalid_i) begin
+          req_state_d = RMW_MODIFY_WRITE;
           mem_valid_o = 1'b1;
           mem_we_o = 1'b1;
           mem_wdata_o = masked_wdata_d;
@@ -157,6 +131,7 @@ module spm_rmw_adapter
         mem_wdata_o = masked_wdata_q;
 
         if (mem_rvalid_i) begin
+          req_state_d = NORMAL;
           mem_valid_o = 1'b0;
           mem_rvalid_o = 1'b1;
         end
@@ -169,7 +144,7 @@ module spm_rmw_adapter
 
   `FF(req_state_q, req_state_d, state_e'('0))
   `FF(mask_q, mask_d, '0)
-  `FF(masked_wdata_q, masked_wdata_d, '0)
+  `FFL(masked_wdata_q, masked_wdata_d, mem_rvalid_i, '0)
   `FF(cnt_q, cnt_d, '0)
 
 endmodule
--- a/hw/ip/spm_interface/src/spm_rmw_adapter.sv
+++ b/hw/ip/spm_interface/src/spm_rmw_adapter.sv
@@ -5,8 +5,10 @@
 // Authors:
 // - Tim Fischer <fischeti@iis.ee.ethz.ch>
 
-`include "common_cells/registers.svh"
+/// This module breaks up sub-word accesses into read-modify-write accesses
+/// to support things like block ECC.
 
+`include "common_cells/registers.svh"
 module spm_rmw_adapter
 #(
   parameter int unsigned  AddrWidth = 0,
@@ -38,28 +40,37 @@ module spm_rmw_adapter
    input logic  mem_ready_i,
    output       addr_t mem_addr_o,
    output       mem_data_t mem_wdata_o,
-   output       mem_strb_t mem_strb_o,
    output logic mem_we_o,
    input logic  mem_rvalid_i,
    input        mem_data_t mem_rdata_i
    );
 
-  typedef enum  {NORMAL, RMW_READ, RMW_MODIFY_WRITE} state_e;
+  typedef enum  {NORMAL, RMW_READ, RMW_MODIFY_WRITE, RMW_FINALIZE} state_e;
+  state_e req_state_q, req_state_d;
 
   mem_data_t mask_q, mask_d;
-  mem_data_t masked_wdata_q, masked_wdata_d, masked_data;
-  assign masked_data = (mem_wdata_i & mask_q) | (mem_rdata_i & ~mask_q);
+  mem_data_t masked_wdata_q, masked_wdata_d;
+  logic masked_wdata_en;
+
+  // The RMW_READ state generates a full-word read request
+  // which is then masked with the pending byte-wise write request
+  assign masked_wdata_d = (mem_wdata_i & mask_q) | (mem_rdata_i & ~mask_q);
+  assign masked_wdata_en = mem_rvalid_i && (req_state_q == RMW_READ);
 
   logic partial_write;
   assign partial_write = mem_valid_i & mem_we_i & ~(&mem_strb_i);
 
-  state_e req_state_q, req_state_d;
-
   cnt_t cnt_q, cnt_d;
   logic rmw_ready, txns_ready;
+
+  // Wait until we do not have any outstanding transactions anymore
+  // since we are blocking the response channel with our `rmw` operation.
   assign rmw_ready = (cnt_q == '0);
+
+  // Only allow `MaxTxns` number of oustanding transactions
   assign txns_ready = cnt_q < MaxTxns;
 
+  // Count number of outstanding requests
   always_comb begin
     cnt_d = cnt_q;
     if (mem_valid_i && mem_ready_o) begin
@@ -74,8 +85,6 @@ module spm_rmw_adapter
     for (int i = 0; i < DataWidth; i++) begin
       mask_d[i] = mem_strb_i[i/8];
     end
-
-    masked_wdata_d = masked_data;
   end
 
   always_comb begin
@@ -84,7 +93,6 @@ module spm_rmw_adapter
     mem_valid_o = mem_valid_i && txns_ready;
     mem_addr_o = mem_addr_i;
     mem_wdata_o = mem_wdata_i;
-    mem_strb_o = '1; // always perform full access
     mem_we_o = mem_we_i;
 
     // Request-side
@@ -94,12 +102,11 @@ module spm_rmw_adapter
 
     req_state_d = req_state_q;
 
-
     unique case (req_state_q)
 
       NORMAL: begin
 
-        // If access is bitwise, generate RMW_READ request
+        // If access is byte-wise, generate full-width read request
         if (partial_write && mem_ready_i && rmw_ready) begin
           req_state_d = RMW_READ;
           mem_we_o = '0;
@@ -108,43 +115,57 @@ module spm_rmw_adapter
 
       RMW_READ: begin
 
+        // stall requests
         mem_rvalid_o = 1'b0;
         mem_rdata_o = '0;
-        mem_we_o = 1'b0;
         mem_ready_o = 1'b0;
 
         if (mem_rvalid_i) begin
           req_state_d = RMW_MODIFY_WRITE;
-          mem_valid_o = 1'b1;
-          mem_we_o = 1'b1;
-          mem_wdata_o = masked_wdata_d;
         end
 
       end // case: RMW_READ
 
       RMW_MODIFY_WRITE: begin
 
+        // stall requests
         mem_rvalid_o = 1'b0;
         mem_ready_o = 1'b0;
 
+        // issue write request with masked data
         mem_valid_o = 1'b1;
         mem_wdata_o = masked_wdata_q;
 
-        if (mem_rvalid_i) begin
-          req_state_d = NORMAL;
-          mem_valid_o = 1'b0;
-          mem_rvalid_o = 1'b1;
+        if (mem_ready_i) begin
+          req_state_d = RMW_FINALIZE;
         end
 
       end // case: RMW_MODIFY_WRITE
+
+      RMW_FINALIZE: begin
+
+        // stall requests
+        mem_rvalid_o = 1'b0;
+        mem_ready_o = 1'b0;
+
+        mem_valid_o = 1'b0;
+        mem_wdata_o = masked_wdata_q;
+
+        // finalize and grant RMW request
+        if (mem_rvalid_i) begin
+          req_state_d = NORMAL;
+          mem_rvalid_o = 1'b1;
+        end
+
+      end // case: RMW_FINALIZE
 
       default: ;
     endcase
   end
 
   `FF(req_state_q, req_state_d, state_e'('0))
-  `FF(mask_q, mask_d, '0)
-  `FFL(masked_wdata_q, masked_wdata_d, mem_rvalid_i, '0)
+  `FFL(mask_q, mask_d, partial_write, '0)
+  `FFL(masked_wdata_q, masked_wdata_d, masked_wdata_en, '0)
   `FF(cnt_q, cnt_d, '0)
 
 endmodule
--- a/hw/ip/spm_interface/src/spm_rmw_adapter.sv
+++ b/hw/ip/spm_interface/src/spm_rmw_adapter.sv
@@ -1,0 +1,225 @@
+// Copyright 2020 ETH Zurich and University of Bologna.
+// Copyright and related rights are licensed under the Solderpad Hardware
+// License, Version 0.51 (the "License"); you may not use this file except in
+// compliance with the License. You may obtain a copy of the License at
+// http://solderpad.org/licenses/SHL-0.51. Unless required by applicable law
+// or agreed to in writing, software, hardware and materials distributed under
+// this License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+// Authors:
+// - Tim Fischer <fischeti@iis.ee.ethz.ch>
+
+module spm_rmw_adapter
+#(
+  parameter int unsigned AddrWidth = 0,
+  parameter int unsigned DataWidth = 0,
+  parameter int unsigned StrbWidth = AddrWidth / 8,
+
+  localparam type addr_t = logic [AddrWidth-1:0],
+  localparam type mem_data_t = logic [DataWidth-1:0],
+  localparam type mem_strb_t = logic [StrbWidth-1:0]
+) (
+   input logic  clk_i,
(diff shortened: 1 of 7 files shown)


hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,225 @@
...
+
+  always_comb begin : next_state_block
+
+    req_state_d = req_state_q;
+
+    unique case (req_state_q)
+      NORMAL: begin
+        req_state_d = NORMAL;
+
+        // If partial memory access (read or write) is detected perform RMW_READ first
```
zarubaf, line 75: I do not think we need to support partial reads. No bus master can actually do such a thing and as the memory has no known side-effect on reads I do not know whether this actually makes sense. So you can assume that this will only be on writes.

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,225 @@
...
+
+  state_t req_state_q, req_state_d;
+
+  always_comb begin : mask_block
+    for (int i = 0; i < DataWidth; i++) begin
+      mask_d[i] = mem_strb_i[i/8];
+    end
+
+    masked_data_d = masked_data_q;
+    if (mem_rvalid_i && !(&mem_strb_i)) begin
```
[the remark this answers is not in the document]
zarubaf: Or, even better, as you are re-using that signal all over the place you probably want to introduce a signal called `partial_write`

```verilog
assign partial_write = mem_rvalid_i & mem_we_i & ~(&mem_strb_i)
```
Imho, it's better to be explicit about the signal and do not rely too much on common sub-expression elimination in the synthesizer.

zarubaf (changes requested):
> The protocol currently takes the following assumptions:
> 
> 1. A valid handshake occurs when `req` and `gnt` are high.

I would probably also prefer if it would be named `valid` and `ready`. Imho that is much more understandable (see the  `stream_to_mem` module).

> 2. The data is read/written one cycle after a valid handshake and the `rvalid` signal is high for both read/write requests

The `rvalid` can come `1..N` cycles after the `reg & gnt` condition. But at least one cycle.

> 3. A `gnt` is raised at the earliest one cycle after the `req` is raised

No `gnt` can come combinatorial in the same cycle (like the `valid`/`ready`).

> 
> Open questions:
> 
> * Are the protocol assumptions correct?

almost

> * Naming and location etc.: the module currently has its own folder `ip/spm_interface`. Could probably be placed somewhere else and maybe named differently. The interface is currently only used for testing purposes

I think eventually that should go into `common_cells`, but having it as an IP is fine for now 👍 

> * Which license header to use?

You can copy the license from existing modules `Solderpad 0.51`.

zarubaf:
What I actually wanted to say is that you can simplify the FSM a bit when not considering partial reads (as indicated by the code comment) - I am expecting that this should very much simplify the FSM. If you want to keep the two FSMs separately that is okay by me but you could also make it one - however you like 😄

hw/ip/spm_interface/src/spm_rmw_adapter.sv
fischeti: I removed support for pariatial reads. The FSM definitely looks much simpler now

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,146 @@
...
+    mem_we_o = mem_we_i;
+
+    // Request-side
+    mem_ready_o = mem_ready_i;
+    mem_rvalid_o = mem_rvalid_i;
+    mem_rdata_o = mem_rdata_i;
+
+    unique case (req_state_q)
+
+      NORMAL: begin
```
zarubaf, line 109: I think there is one edge case that isn't yet handled. Theoretically, the protocol allows for an infinite amount of outstanding transactions we probably need to keep a counter to whether there are any outstanding otherwise we are potentially suppressing the answer of any outstanding transaction returning a datum.

So a counter that increments whenever a new transaction is accepted and decrements whenever it finishes should be sufficient. If it is `0` we can go into the next RMW state. We also need to make sure that we do not accept any in case the counter would overflow.

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,147 @@
...
+  end
+
+  always_comb begin
+
+    req_state_d = req_state_q;
+
+    unique case (req_state_q)
+      NORMAL: begin
+        // If partial memory access is detected perform RMW_READ
+        if (partial_write) begin
```
zarubaf, line 70: Like this, right? Otherwise, we would switch state even though the memory didn't accept yet.
```suggestion
        if (partial_write && mem_ready_i) begin
```

zarubaf (changes requested):

hw/ip/spm_interface/src/spm_rmw_adapter.sv
fischeti: Yes, you are right of course.

hw/ip/spm_interface/src/spm_rmw_adapter.sv
fischeti: Makes sense. I have added `cnt_q` that prevents new transactions if is equal a configurable`MaxTxns` and stalls RMW requests if not equal `0`.

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,175 @@
...
+      RMW_MODIFY_WRITE: begin
+        if (mem_rvalid_i) begin
+          req_state_d = NORMAL;
+        end
+      end
+
+      default: ;
+
+    endcase
+  end
```
zarubaf, line 110: At this stage, I think the code would be more readable if the two FSMs would be merged.

zarubaf (commented):
Almost good to go, let's unify the two FSM - that makes it less code and easier to reason imho.

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,175 @@
...
+  logic         rmw_ready, txns_ready;
+  assign rmw_ready = (cnt_q == '0);
+  assign txns_ready = cnt_q < MaxTxns;
+
+  always_comb begin
+    cnt_d = cnt_q;
+    if (mem_valid_o && mem_ready_i) begin
+      cnt_d++;
+    end
+     if (mem_rvalid_i) begin
```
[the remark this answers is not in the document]
fischeti: The RMW operation starts only if the counter is `0` and should only have one outstanding transaction at any moment. But I will change it, it is probably a bit safer that way.

hw/ip/spm_interface/src/spm_rmw_adapter.sv
fischeti: Yes, I agree I will merge it.

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,150 @@
...
+// Solderpad Hardware License, Version 0.51, see LICENSE for details.
+// SPDX-License-Identifier: SHL-0.51
+
+// Authors:
+// - Tim Fischer <fischeti@iis.ee.ethz.ch>
+
+`include "common_cells/registers.svh"
+
+module spm_rmw_adapter
+#(
```
zarubaf, line 11: Nitpicking, but a small comment what the module does would be quite good. That helps people understand the intent i.e., `/// This module breaks up sub-word accesses into read-modify-write accesses to support things like block ECC.`

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,150 @@
...
+   input logic  mem_we_i,
+   output logic mem_rvalid_o,
+   output       mem_data_t mem_rdata_o,
+
+   // Mem-side channel
+   output logic mem_valid_o,
+   input logic  mem_ready_i,
+   output       addr_t mem_addr_o,
+   output       mem_data_t mem_wdata_o,
+   output       mem_strb_t mem_strb_o,
```
zarubaf, line 41: I think to avoid confusion we can remove this port entirely.

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,150 @@
...
+  assign masked_data = (mem_wdata_i & mask_q) | (mem_rdata_i & ~mask_q);
+
+  logic partial_write;
+  assign partial_write = mem_valid_i & mem_we_i & ~(&mem_strb_i);
+
+  state_e req_state_q, req_state_d;
+
+  cnt_t cnt_q, cnt_d;
+  logic rmw_ready, txns_ready;
+  assign rmw_ready = (cnt_q == '0);
```
zarubaf, line 60:
```suggestion
  // Wait until we do not have any outstanding transactions anymore
  // since we are blocking the response channel with our `rmw` operation.
  assign rmw_ready = (cnt_q == '0);
```

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,150 @@
...
+
+      end // case: RMW_READ
+
+      RMW_MODIFY_WRITE: begin
+
+        mem_rvalid_o = 1'b0;
+        mem_ready_o = 1'b0;
+
+        mem_valid_o = 1'b1;
+        mem_wdata_o = masked_wdata_q;
```
zarubaf, line 131: I think we have a protocol violation here, don't we? By keeping `mem_valid_o` asserted you are triggering potentially many more outstanding write transactions. Unfortunately, I do not see how we could do that without introducing another state. We would transition to `RMW_FINALIZE` when we get a `mem_ready_i`. And then from finalize to `NORMAL` on `mem_rvalid_i`. Does that make sense?

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,150 @@
...
+        mem_rdata_o = '0;
+        mem_we_o = 1'b0;
+        mem_ready_o = 1'b0;
+
+        if (mem_rvalid_i) begin
+          req_state_d = RMW_MODIFY_WRITE;
+          mem_valid_o = 1'b1;
+          mem_we_o = 1'b1;
+          mem_wdata_o = masked_wdata_d;
+        end
```
zarubaf, line 121: Potentially saves a cycle but introduces quite a long path between memory read and write. I would remove that (also makes it a bit simpler to reason about). I don't think that sub-word accesses on this memory are really a performance bottleneck. Do you agree?

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,150 @@
...
+
+      end // case: RMW_MODIFY_WRITE
+
+      default: ;
+    endcase
+  end
+
+  `FF(req_state_q, req_state_d, state_e'('0))
+  `FF(mask_q, mask_d, '0)
+  `FFL(masked_wdata_q, masked_wdata_d, mem_rvalid_i, '0)
```
zarubaf, line 147: Hypothetically speaking we could condition the load condition also on the state (would help the synthesiser with inferring a clock gate).

hw/ip/spm_interface/src/spm_rmw_adapter.sv
fischeti: Yes, I haven't thought about the long path. Also, I think the FSM would look more comprehensible by removing it. I was not sure how much sacrificing a cycle impacts performance, but if this negligible than I totally agree on changing this.

hw/ip/spm_interface/src/spm_rmw_adapter.sv
fischeti: Yes that is actually true. I will add an additional state `RMW_FINALIZE` in that case.

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,171 @@
...
+  always_comb begin
+
+    // Mem-side
+    mem_valid_o = mem_valid_i && txns_ready;
+    mem_addr_o = mem_addr_i;
+    mem_wdata_o = mem_wdata_i;
+    mem_we_o = mem_we_i;
+
+    // Request-side
+    mem_ready_o = mem_ready_i && txns_ready && rmw_ready;
```
zarubaf, line 99: We are unnecessarily stalling if the operation isn't `rmw`, right? That would be a fix:
```suggestion
    mem_ready_o = mem_ready_i & txns_ready & !partial_write;
```

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,171 @@
...
+  always_comb begin
+    for (int i = 0; i < DataWidth; i++) begin
+      mask_d[i] = mem_strb_i[i/8];
+    end
+  end
+
+  always_comb begin
+
+    // Mem-side
+    mem_valid_o = mem_valid_i && txns_ready;
```
zarubaf, line 93: Nit (afaik from the style guide):
```suggestion
    mem_valid_o = mem_valid_i & txns_ready;
```

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,171 @@
...
+          req_state_d = RMW_READ;
+          mem_we_o = '0;
+        end
+      end // case: NORMAL
+
+      RMW_READ: begin
+
+        // stall requests
+        mem_rvalid_o = 1'b0;
+        mem_rdata_o = '0;
```
zarubaf, line 120: Not necessary, right? We are just here when we do a read, so the `rdata` field will not carry any meaningful payload (saves some gates).
```suggestion
```

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,171 @@
...
+        if (mem_ready_i) begin
+          req_state_d = RMW_FINALIZE;
+        end
+
+      end // case: RMW_MODIFY_WRITE
+
+      RMW_FINALIZE: begin
+
+        // stall requests
+        mem_rvalid_o = 1'b0;
```
zarubaf, line 148:
```suggestion
        mem_rvalid_o = mem_rvalid_i;
```

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,171 @@
...
+        mem_ready_o = 1'b0;
+
+        mem_valid_o = 1'b0;
+        mem_wdata_o = masked_wdata_q;
+
+        // finalize and grant RMW request
+        if (mem_rvalid_i) begin
+          req_state_d = NORMAL;
+          mem_rvalid_o = 1'b1;
```
zarubaf, line 157:
```suggestion
```

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,171 @@
...
+        end
+
+      end // case: RMW_FINALIZE
+
+      default: ;
+    endcase
+  end
+
+  `FF(req_state_q, req_state_d, state_e'('0))
+  `FFL(mask_q, mask_d, partial_write, '0)
```
zarubaf, line 167: We could save the `mask_q` register here if we stall the incoming transaction until `masked_wdata_en`, right? Then we would still have `mask` available and generate the `masked_wdata_q` content.
```suggestion
```

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,171 @@
...
+          mem_we_o = '0;
+        end
+      end // case: NORMAL
+
+      RMW_READ: begin
+
+        // stall requests
+        mem_rvalid_o = 1'b0;
+        mem_rdata_o = '0;
+        mem_ready_o = 1'b0;
```
zarubaf, line 121:
```suggestion
        mem_ready_o = mem_rvalid_i;
```

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,171 @@
...
+  state_e req_state_q, req_state_d;
+
+  mem_data_t mask_q, mask_d;
+  mem_data_t masked_wdata_q, masked_wdata_d;
+  logic masked_wdata_en;
+
+  // The RMW_READ state generates a full-word read request
+  // which is then masked with the pending byte-wise write request
+  assign masked_wdata_d = (mem_wdata_i & mask_q) | (mem_rdata_i & ~mask_q);
+  assign masked_wdata_en = mem_rvalid_i && (req_state_q == RMW_READ);
```
zarubaf, line 58:
```suggestion
```

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,171 @@
...
+      cnt_d++;
+    end
+     if (mem_rvalid_o) begin
+       cnt_d--;
+     end
+  end
+
+  always_comb begin
+    for (int i = 0; i < DataWidth; i++) begin
+      mask_d[i] = mem_strb_i[i/8];
```
zarubaf, line 86:
```suggestion
      mask[i] = mem_strb_i[i/8];
```

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,171 @@
...
+  typedef enum  {NORMAL, RMW_READ, RMW_MODIFY_WRITE, RMW_FINALIZE} state_e;
+  state_e req_state_q, req_state_d;
+
+  mem_data_t mask_q, mask_d;
+  mem_data_t masked_wdata_q, masked_wdata_d;
+  logic masked_wdata_en;
+
+  // The RMW_READ state generates a full-word read request
+  // which is then masked with the pending byte-wise write request
+  assign masked_wdata_d = (mem_wdata_i & mask_q) | (mem_rdata_i & ~mask_q);
```
zarubaf, line 57:
```suggestion
  assign masked_wdata_d = (mem_wdata_i & mask) | (mem_rdata_i & ~mask);
```

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,171 @@
...
+   output       mem_data_t mem_wdata_o,
+   output logic mem_we_o,
+   input logic  mem_rvalid_i,
+   input        mem_data_t mem_rdata_i
+   );
+
+  typedef enum  {NORMAL, RMW_READ, RMW_MODIFY_WRITE, RMW_FINALIZE} state_e;
+  state_e req_state_q, req_state_d;
+
+  mem_data_t mask_q, mask_d;
```
zarubaf, line 51:
```suggestion
  mem_data_t mask;
```

zarubaf (commented):
Okay, sorry for yet another review - There is one minor modification that I thought would be nice: Removign the `mask_q` register by delaying the partial write transaction until we can directly save the masked data. Potentially saves quite some flip-flops. The rest is nits. I've annotated the change with change requests but of course, I am not sure if it is entirely correct.

hw/ip/spm_interface/src/spm_rmw_adapter.sv
fischeti: This fix causes the `mem_ready_o` to never be asserted for rmw requests, so I have slightly modified it:
```suggestion
    mem_ready_o = mem_ready_i & txns_ready & !(partial_write & !rmw_ready);
```

hw/ip/spm_interface/src/spm_rmw_adapter.sv
fischeti: Okay, I more or less follow [this styleguide](https://github.com/lowRISC/style-guides/blob/998f5a1f9146825b35bab87f944acca2424f90f3/VerilogCodingStyle.md#logical-vs-bitwise) which suggests logical `&&` operators for logic operations. But I have noticed that the existing codebase mostly follows the bitwise convention, so I will bring my code in line😄

hw/ip/spm_interface/src/spm_rmw_adapter.sv
fischeti: Yes, this should be possible. I am actually not sure why I initially instanstiated registers for the `mask`. But under the assumption that the `mem_strb_i` remains stable when stalling the incoming transaction, we can remove it.

hw/ip/spm_interface/src/spm_rmw_adapter.sv
fischeti: I don't quite understand this. We still need a enable signal for `masked_wdata_q`, right?

hw/ip/spm_interface/src/spm_rmw_adapter.sv
```diff
@@ -0,0 +1,168 @@
...
+  always_comb begin
+
+    // Mem-side
+    mem_valid_o = mem_valid_i & txns_ready;
+    mem_addr_o = mem_addr_i;
+    mem_wdata_o = mem_wdata_i;
+    mem_we_o = mem_we_i;
+
+    // Request-side
+    mem_ready_o = mem_ready_i & txns_ready & !(partial_write & !rmw_ready);
```
zarubaf, line 99: Somehow I have trouble reasoning over this. Shouldn't that be? I might also be wrong about this.
```suggestion
    mem_ready_o = mem_ready_i & txns_ready & !partial_write;
```

fischeti: https://github.com/pulp-platform/snitch/pull/164#discussion_r657112369
`partial_write` is asserted for RMW writes until a handshake occurs. But a handshake will not happen because `mem_ready_o` is never asserted due to `!partial_write`. Does that make sense, or did I miss anything?

zarubaf: Right, but you are overwriting `mem_ready_o` in `RMW_READ`, right - so that shouldn't be a problem? You do not want to handshake before that for RMW operations or otherwise we are loosing the byte-mask, aren't we?

fischeti: Yes you are actually correct. I think the issue I explained occured when `mem_ready_o` was not yet overwritten in `RMW_READ`. But in that case it is not needed anymore.

fischeti:
I have actually discovered another protocol violation. `mem_ready_o` should not be asserted in the `RMW_READ` state but it should be delayed until the `mem_ready_i` is asserted in the `RMW_MODIFY_WRITE` state. Otherwise we lose the address.

zarubaf (approved):
Good catch, LGTM to me now!

PR #166, opened 2021-06-21
solder: Fix AXI serialize template

stmach:


zarubaf:
El clássico. Thanks for fixing.

PR #168, opened 2021-06-24
Occamy: Increase cacheline size to 256 bit

fischeti:
FPU utilization is limited by instruction cache misses in some cases. Increasing the cacheline from 128bit to 256bit mitigates this issues to some extent.


zarubaf (approved):
Thanks!

zarubaf (approved):

PR #169, opened 2021-06-25
banshee: SSR num streamers to 3 and almost parametric

noah95:
This changes the number of SSR streamers from 2 to 3 as is the default in `system/snitch_cluster` and **almost** makes the number of SSRs parametric. The CPU struct is still hard-coded to 3 SSRs. 

## ToDo
- [x] Make the CPU struct parametric to `SSR_N_STREAMERS`
- [x] Pass `SSR_N_STREAMERS` in the banshee config yml


noah95:
@paulsc96 could you check why the ISSR tests fail? Might there be anything hard-coded in the indirection extension  that I missed?

SamuelRiedel (approved):
LGTM. Thanks :+1:

PR #170, opened 2021-06-28
Fix: peripheral counters

fischeti:
Fixes some bugs for configuring the snitch cluster peripheral performance counters.


zarubaf (approved):
Thanks, LGTM!

PR #171, opened 2021-06-30
ssr: Fix register interface to maximum of 4 loops

paulsc96:
This change fixes the register layout for iteration loops in the SSR configuration interface to 4 loops, even if less loops are instantiated; this ensures a consistent configuration layout no matter the parameterization.

The hardware itself remains parameterizable in the same way as before.


zarubaf (approved):
LGTM

PR #172, opened 2021-06-30
ssr: Fix shadowing bug in indirector initialization

paulsc96:
This fixes a bug wherein the SSR indirector is not correctly initialized when a shadowed configuration is written while another job is still active.


zarubaf (approved):
That sounds like a nasty bug :(

PR #174, opened 2021-07-01
Peripherals added using callback functions

jeudine:
To create your own custom peripherals, just modify the user part in `src/periphs.rs`

### DIFF
--- a/sw/banshee/config/mempool.yaml
+++ b/sw/banshee/config/mempool.yaml
@@ -9,8 +9,6 @@ address:
   tcdm_start: 0x40000008
   tcdm_end: 0x4000000C
   nr_cores: 0x40000010
-  periph_start: 0x40000014
-  periph_end: 0x40000018
   uart: 0xC0000000
   # Not supported in MemPool
   barrier_reg: 0x50000000
--- a/sw/banshee/config/mempool.yaml
+++ b/sw/banshee/config/mempool.yaml
@@ -22,26 +22,19 @@ memory:
     start: 0x0
     end: 0x100000
     latency: 5
-  periph:
+  periphs:
     start: 0x100000
     end: 0x200000
     latency: 5
+    func:
         - name: PERF_COUNTER_ENABLE
-          access: rw
           size: 6
-          fn: perf_counter_enable
         - name: HART_SELECT
-          access: rw
           size: 10
-          fn: hart_select
         - name: PERF_COUNTER
-          access: rw
           size: 48
-          fn: perf_counter
         - name: WAKE_UP
-          access: wo
           size: 32
-          fn: wake_up
   dram:
     start: 0x80000000
     end: 0x80010000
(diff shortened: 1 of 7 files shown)


sw/banshee/src/engine.rs
```diff
@@ -326,6 +325,13 @@ impl Engine {
             (0..self.num_clusters).map(|_| tcdm.clone()).collect()
         };
 
+        // Allocate the pripherals
+        let peripherals: Vec<_> = {
+            (0..self.num_clusters)
+                .map(|_| Peripherals::new(&self.config.memory.periphs.callbacks))
+                .collect()
+        };
+
```
SamuelRiedel, line 334: This is great for now. For Occamy, we will need the same mechanism as for the TCDM, i.e., each cluster has its own peripherals and tcdm, but with different addresses, and the clusters need access to one another's tcdm and peripherals. But that can be tackled in the next step. @paulsc96 Does that also hold for the 'vanilla' Snitch cluster?

sw/banshee/doc/peripheral.adoc
```diff
@@ -0,0 +1,123 @@
+= Peripherals
```
SamuelRiedel, line 1: I like it :+1:

sw/banshee/config/mempool.yaml
```diff
@@ -20,6 +20,13 @@ memory:
     start: 0x0
     end: 0x100000
     latency: 5
+  periphs:
```
SamuelRiedel, line 23: Did you use this config for your tests? I don't think it makes sense to modify the MemPool config, but rather create a new configuration for your tests.

SamuelRiedel (commented):
Looks mostly good to me. It would be nice if you could add some tests. I am sure you have some tests which helped you develop.

sw/banshee/src/engine.rs
paulsc96: There is no difference between an Occamy Snitch cluster and a 'vanilla' one except parameterization. So yes, all cluster-local resources should be reachable from anywhere externally and *not* alias with anything else if configured accordingly.

This being said, there could be a usecase for system-wide peripherals. But we should focus on our goal here, which is fully functional multi-cluster emulation.

jeudine:
I added a test in the last commit, which verify that the example peripheral (peripheral access counter) is working.
In the end the test result is passed for all the tests
I also reseted the MemPool config, and created a new test file just for the example peripheral

sw/banshee/config/mempool.yaml
jeudine: I reset the MemPool config file and added a new one for the test

noah95:
@jeudine could you rebase onto master and resolve the conflict? We can then enable the CI to check if anything breaks

jeudine:
> @jeudine could you rebase onto master and resolve the conflict? We can then enable the CI to check if anything breaks

Conflict resolved @noah95

zarubaf:
@jeudine Sorry, I think you need to merge `master` once more. #202 has just been merged now.

PR #175, opened 2021-07-05
banshee: Modify `num_sleep` only by the hart that enters/leaves wfi

noah95:
I suggest changing the `num_sleep` member only when entering/leaving the sleep state. The current implementation can cause underflow in `num_sleep` if a core is woken that is not at `wfi` resulting in `banshee` not exiting although all harts are sleeping. This also better reflects the hardware's behavior

### DIFF
--- a/sw/banshee/src/engine.rs
+++ b/sw/banshee/src/engine.rs
@@ -717,7 +717,12 @@ impl<'a, 'b> Cpu<'a, 'b> {
         let hartid = self.hartid - self.engine.base_hartid;
         while self.wake_up[hartid].load(Ordering::Relaxed) == 0 {
             // Check if everyone is sleeping
-            if self.num_sleep.load(Ordering::Relaxed) == self.wake_up.len() {
+            if self.num_sleep.load(Ordering::Relaxed) == self.wake_up.len()
+                && self
+                    .wake_up
+                    .into_iter()
+                    .all(|x| x.load(Ordering::Relaxed) == 0)
+            {
                 return 1;
             }
             std::thread::yield_now();
--- a/sw/banshee/src/engine.rs
+++ b/sw/banshee/src/engine.rs
@@ -553,13 +553,10 @@ impl<'a, 'b> Cpu<'a, 'b> {
             x if x == self.engine.config.address.wakeup_reg => {
                 // wake_up
                 if value as i32 == -1 {
-                    self.num_sleep
-                        .fetch_sub(self.wake_up.len(), Ordering::Relaxed);
                     for x in self.wake_up {
                         x.fetch_max(self.state.cycle + 1, Ordering::Release);
                     }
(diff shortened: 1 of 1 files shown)


SamuelRiedel:
Good point. The problem currently (and the reason it is implemented as it was) is that we can now have a race condition terminating some threads even if others are still active.

In the scenario where only one core is awake, it triggers the next core to wake up, and then goes to sleep itself, it can be that the `num_sleep` counter is equal to the number of cores, and cores that stay asleep can already terminate.

We could just add the additional constraint that the `wake_up` array must be empty for a thread to terminate. I think that would eliminate the mentioned race condition.

SamuelRiedel:
I think the CI is stuck as well... Let me quickly check what the `wfi` test does.

SamuelRiedel:
The problem is a tricky race condition. Basically what the threads now do is the following:
```
1. num_sleep++        // From no on, we risk terminating if there is an outstanding wakeup
2. check(num_sleep)
3. check(wake_up)     // We yield if both checks are satisfied
4. wake_up[x] = 0
5. num_sleep--
```
The problem is that currently 2 & 3 are not atomic. Considering the following interleaving,:
```
Start: wake_up[B] = 1;
Thread A: 1 2    3 --> wake_up is all zeros, so terminate
Thread B:  1 2345  --> woke up
```
I don't think we can do this with two variables, so back to the drawing board. In which scenarios did you encounter an underflow of `num_sleep`, which lead to the simulation not stopping?

noah95:
> I don't think we can do this with two variables, so back to the drawing board. In which scenarios did you encounter an underflow of `num_sleep`, which lead to the simulation not stopping?

In my application I wake the DMA core to perform a task. If the next task is being scheduled before the previous is complete, the DMA core is not yet in `wfi` and `num_sleep` gets decremented by the second `wake_up`. So perhaps we could check if the core to be woken is at sleep and skip the `wake_up` if it is already running?

SamuelRiedel:
> So perhaps we could check if the core to be woken is at sleep and skip the `wake_up` if it is already running?

I see, is this how it's implemented in hardware at the moment?

I can still see race conditions occurring in this scenario. The DMA core could check whether a new task is available which is not the case, so it will go to sleep in the next step. In the meantime, the other core pushes a new task to the queue and sends the wake-up trigger, which will be ignored however since the DMA core is still awake, even though it's about to shut down.

Maybe we should consider using the official RISC-V interrupts, instead of our own 'wake-up' triggers. In that case, the core could clear its interrupt register without going to sleep.

noah95:
> Maybe we should consider using the official RISC-V interrupts, instead of our own 'wake-up' triggers. In that case, the core could clear its interrupt register without going to sleep.

This is a far more sophisticated approach for this application, true. Let me close this for now and look into interrupts, i think there has already been work done in RTL that we could port to banshee.

PR #177, opened 2021-07-14
occamy: Standalone boot flow with U-Boot SPL

niwis:
This integrates U-Boot SPL into the bootrom of Occamy on FPGA. Furthermore, it adds Vivado scripts for loading U-Boot and Linux into the VCU128's SPI flash.


zarubaf (approved):
Terrific and perfect! Thanks!

Issue #176, opened 2021-07-09
Multicluster DMA transfer deadlock


fischeti:
Simulatneously copying or sharing data across clusters in a multicluster system (i.e. occamy) can cause deadlocks. The DMA of each cluster issues a read request to another clusters TCDM and a corresponding write request to its own TCMD. Hence each cluster has an incoming read request and a local write request it needs to arbitrate which is handled in the `axi_to_mem` module. https://github.com/pulp-platform/snitch/blob/ca929ad402878b64ab3746d86fb25ae6d245930f/hw/ip/future/src/axi_to_mem.sv#L217-L259
The module currently arbitrates in round-robin fashion (among others prioritizations) which might cause multiple clusters to simultaneously service the incoming read request and stall the local write request. This situation causes a deadlock since the read request depends on the write request to be serviced. 

A current workaround for this issue is to naively prioritize write over read requests. But this only fixes this particular problem and might cause different deadlocks.

PR #178, opened 2021-07-26
Update axi_to_mem module to prevent deadlock

fischeti:
This PR addresses  #176. The new `axi_to_mem` module prevents deadlocks by allowing reads to bypass writes and vice versa and should not have a negative impact on performance. It is not 100% guaranteed that this will solve all deadlock cases, hence support for serialization should be added in the future

### DIFF
--- a/hw/ip/future/src/axi_to_mem_interleaved.sv
+++ b/hw/ip/future/src/axi_to_mem_interleaved.sv
@@ -1,0 +1,216 @@
+// Copyright 2018 ETH Zurich and University of Bologna.
+// Copyright and related rights are licensed under the Solderpad Hardware
+// License, Version 0.51 (the "License"); you may not use this file except in
+// compliance with the License.  You may obtain a copy of the License at
+// http://solderpad.org/licenses/SHL-0.51. Unless required by applicable law
+// or agreed to in writing, software, hardware and materials distributed under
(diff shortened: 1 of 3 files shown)


hw/ip/future/src/axi_to_mem_interleaved.sv
```diff
@@ -0,0 +1,210 @@
+// Copyright 2020 ETH Zurich and University of Bologna.
+// Solderpad Hardware License, Version 0.51, see LICENSE for details.
+// SPDX-License-Identifier: SHL-0.51
+
+// Author:
+// Thomas Benz <tbenz@iis.ee.ethz.ch>
+
+/// axi_to_mem_interleved prevents deadlocks by allowing reads to
```
zarubaf, line 8:
```suggestion
/// `axi_to_mem_interleaved` prevents deadlocks by allowing reads to
```

zarubaf (approved):
LGTM! Thanks

hw/ip/future/src/axi_to_mem_interleaved.sv
```diff
@@ -0,0 +1,210 @@
+// Copyright 2020 ETH Zurich and University of Bologna.
```
zarubaf, line 1:
```suggestion
// Copyright 2021 ETH Zurich and University of Bologna.
```

Issue #167, opened 2021-06-21
Register Write assertion on register x0


noah95:
This assertion should not trigger when writing to `x0`

https://github.com/pulp-platform/snitch/blob/6f7ed31cdade5b68c74cb658ebf9ed859ef85e56/hw/ip/snitch/src/snitch.sv#L2606

zarubaf:
That should be fixed by #180

PR #180, opened 2021-07-30
hw: Disable assertion on write to x0

noah95:
This addresses issue #167 where a write to `x0` causes the assertion to fail

### DIFF
--- a/hw/ip/snitch/src/snitch.sv
+++ b/hw/ip/snitch/src/snitch.sv
@@ -2603,6 +2603,6 @@ module snitch import snitch_pkg::*; import riscv_instr::*; #(
       (inst_valid_o && inst_ready_i && inst_cacheable_o) ##1 (inst_valid_o && $stable(inst_addr_o))
       |-> inst_ready_i && $stable(inst_data_i), clk_i, rst_i)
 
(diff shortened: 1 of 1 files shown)


hw/ip/snitch/src/snitch.sv
```diff
@@ -2603,6 +2603,6 @@ module snitch import snitch_pkg::*; import riscv_instr::*; #(
       (inst_valid_o && inst_ready_i && inst_cacheable_o) ##1 (inst_valid_o && $stable(inst_addr_o))
       |-> inst_ready_i && $stable(inst_data_i), clk_i, rst_i)
 
-  `ASSERT(RegWriteKnown, gpr_we |-> !$isunknown(gpr_wdata), clk_i, rst_i)
+  `ASSERT(RegWriteKnown, gpr_we & gpr_waddr |-> !$isunknown(gpr_wdata), clk_i, rst_i)
```
zarubaf, line 2606: I don't think that is entirely correct, right?
```suggestion
  `ASSERT(RegWriteKnown, gpr_we & (gpr_waddr != 0) |-> !$isunknown(gpr_wdata), clk_i, rst_i)
```

noah95 (commented):
Of course, that is correct thanks

PR #184, opened 2021-08-02
occamy: Add isolation signals to S1 quadrants

zarubaf:
Since the clusters will start issuing AXI transactions when they boot, we want to safeguard AXI transactions since - in case a given cluster isn't performing correctly - lock up the entire system.

Isolation is controlled via the SoC control register in the top level.


niwis:
Looks good to me, only one question: why are 12 isolate registers defined? I can only see 8 being used in `occamy_top`

zarubaf:
That's right, my laziness in over-provisioning them and just in case we end up with more quadrants then 8 at the moment. I think in a perfect world we would generate the `hjson` file from the Occamy description. Do you think we can merge with that hack?

niwis (approved):
Sure, makes sense! 👍

PR #186, opened 2021-08-04
occamy: Hook-up interrupts

zarubaf:
This add timer interrupts and software interrupts to each snitch core. Imho, up for discussion should be whether we are really want to have a `timercmp` for each Snitch core. We can definitely customize this.


niwis (approved):
Nice. That looks like one large CLINT ;-)

zarubaf:
Hehe, yes - it is :)

PR #187, opened 2021-08-04
Add atomic adapter for SPM

zarubaf:
As discussed (@niwis) it would be good for the SPM to support atomics so that we can properly use U-Boot SPL etc. from the SPM.


niwis (changes requested):
Great!! I guess #186 has sneaked in?

zarubaf:
Yes, re-based on latest master. The diff should be cleaner now.

niwis (approved):
Perfect, looks good to me!

PR #188, opened 2021-08-05
banshee: Default TCDM address same as cluster

noah95:
Why don't we default to the same TCDM start address as the `snitch_cluster`? While adding an allocator to the `snRuntime` the first allocation is currently at `0x0` which is a null pointer signalling a failed allocation. Let's see if this brakes any CI.


SamuelRiedel:
Absolutely. It seems the multicluster tests currently fail. Most likely because the tests are not recompiled with the updated TCDM addresses, which leads to the TCDM being mapped to the slow path and be shared amongst all clusters. I think once you align the linker script and recompile the tests, most of them should already pass.

SamuelRiedel (approved):
Ok, this seems to have fixed all the tests :tada: 

LGTM Thanks :)

PR #190, opened 2021-08-06
occamy: Fix SPM PMA to executable and cacheable

zarubaf:
As briefly discussed this fixes the PMA region of CVA6 for the SPM to be cacheable and executable so that we can properly boot from it.


niwis (approved):
Looks good to me! I will adapt the boot flow.

PR #191, opened 2021-08-06
icache: Fix refiller in case of larger AXI port

zarubaf:
In case the AXI port is larger than the cache line the refiller would
violate the protocol by using the address to re-align the incoming R
beat. But the address wasn't stable anymore since the transaction had
already handshaked.

The only solution would be to safe the offset and re-align when the Rs
are coming in.


zarubaf:
I think that is pretty uncontroversial - so I am going to merge it :)

PR #192, opened 2021-08-06
Instruction cache at wide cluster port

zarubaf:
Now that we have the wide DMA port it (can) make sense to actually
perform icache refills from that port, this potentially cuts down the
re-fill latency a couple of cycles.

More importantly, in case there is a read-only cache on the wide
network, it makes use of the same.

(Based on #191)

### DIFF
--- a/hw/ip/snitch/src/snitch_pkg.sv
+++ b/hw/ip/snitch/src/snitch_pkg.sv
@@ -128,14 +128,14 @@ package snitch_pkg;
   typedef enum integer {
     CoreReq = 0,
     AXISoC  = 1,
-    PTW     = 2,
-    ICache  = 3
+    PTW     = 2
   } cluster_master_e;
 
   // Slaves on Cluster DMA AXI Bus
   typedef enum int unsigned {
-    TCDMDMA = 32'd0,
-    SoCDMAOut = 32'd1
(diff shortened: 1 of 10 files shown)


zarubaf:
@niwis Anything you want to add/change?

niwis (approved):
Looks good to me! My only concern is that instruction fences remain effective, what is the solution there? Flushing the full quadrant? (Sorry for the delayed reply, the PR slipped my attention)

niwis:
And is it possible to access the SPM via the wide crossbar?

zarubaf:
> Looks good to me! My only concern is that instruction fences remain effective, what is the solution there? Flushing the full quadrant? (Sorry for the delayed reply, the PR slipped my attention)

I don't think anything should change for that. The instruction cache can still be invalidated very much the same way (`fence.i`).



> And is it possible to access the SPM via the wide crossbar?

That is a very valid point. Currently, it can't, but it seems that this needs to change anyway (see the discussion yesterday on the atomics supported in L2). I will open a new issue for that, it should be relatively easy to fix (famous last words).

niwis:
Okay :-) And I guess the same then also goes for the bootrom, debug module and other peripherals

niwis:
Looks good to me after #200

PR #193, opened 2021-08-06
Feature/cluster interrupts

noah95:


sw/snRuntime/src/start.S
```diff
@@ -44,6 +44,45 @@ snrt.crt0.init_bss:
     blt       t0, t1, 1b
 2:
 
+snrt.crt0.init_registers:
+    # Clear FP registers
+    csrr    t0, misa
+    andi    t0, t0, (1 << 3) # 3: D - single precision float extension
```
zarubaf, line 50: What about F-only cores? I think we should also use `f` instructions here, since they will also clear the register.

We could also set `mtvec` to the `3f` label. In case the core doesn't have the available instructions it should except to that address with an illegal instruction.

sw/banshee/src/runtime/common.rs
```diff
@@ -81,3 +83,23 @@ pub struct DmaState {
...
+    pub mie: u32,
+    // machine interrupt pending (same bits as mip csr)
+    pub mip: u32,
+    // machine trap vector
+    pub mtvec: u32,
+    // machine exception PC
+    pub mepc: u32,
+    // machine cause
+    pub mcause: u32,
+}
```
zarubaf, line 105: The only real comment I would have is that this seems a bit ad-hoc to me, I can see how `mcause` relates to interrupts but I would probably prefer if we can group the CSRs in something like a privileged state (in case we are extending the capabilities later). But I am okay with leaving it like this now.

zarubaf (approved):
Generally LGTM! Thanks a lot for this very nice addition!

hw/ip/snitch_cluster/src/snitch_cluster_wrapper.sv.tpl
```diff
@@ -11,21 +11,21 @@ ${disclaimer}
...
-  % for c in reversed(cfg['cores']):
+  % for c in cfg['cores'][::-1]:
 ${int(c[prop])}\
   % endfor
 </%def>\
 
 <%def name="core_isa(isa)">\
 ${cfg['nr_cores']}'b\
-  % for c in cfg['cores']:
+  % for c in reversed(cfg['cores']):
```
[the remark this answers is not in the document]
noah95: `reversed` didnt work [here](https://github.com/pulp-platform/snitch/pull/193/files/2556c294cb8964ec9458aa5bb982b7ad43c5fca8#diff-c91009ca628484eaed6d84ddeef3791d8f5c928e87619d49d8909af339283f6cR14) so this is just an artifact and is now resolved

sw/snRuntime/src/start.S
noah95: Absolutely correct but I wouldn't do it with an exception. Resolved in 5d61898

sw/banshee/src/runtime/common.rs
noah95: Agree, we could simply group all CSRs into a HashMap for example. @SamuelRiedel any suggetions?

SamuelRiedel: A hash map sounds like a good idea. But some special registers like `minstret`, which is accessed each instruction, might better be kept directly accessible.

sw/banshee/src/engine.rs
```diff
@@ -647,23 +703,74 @@ impl<'a, 'b> Cpu<'a, 'b> {
...
+            riscv::Csr::Mie => self.state.irq.mie = value,         // CSR_MIE
+            riscv::Csr::Mip => self.state.irq.mip = value,         // CSR_MIP
+            riscv::Csr::Mtvec => self.state.irq.mtvec = value,     // CSR_MTVEC
+            riscv::Csr::Mepc => self.state.irq.mepc = value,       // CSR_MEPC
+            riscv::Csr::Mcause => self.state.irq.mcause = value,   // CSR_MCAUSE
+            _ => (),
+        }
+    }
+
+    fn binary_csr_read_silent(&self, csr: riscv::Csr) -> u32 {
```
SamuelRiedel, line 743: Is the only difference to the 'regular' car functions the tracing?

SamuelRiedel (approved):
LGTM. Thanks :) I think the solution for the IRQ checking is fine and we can always improve it when necessary. I just have some minor questions.

sw/banshee/src/engine.rs
noah95: yes the `trace!` is the only difference. Since with the interrupts the CSR are read on every instruction this would else pollute the output when running with `SNITCH_LOG=trace` (has nothing todo with the `--trace` tracer).

The redundant code is ugly, should I solve this with a function argument?

PR #194, opened 2021-08-09
Fix CSR wfi

noah95:
This specific sequence of instructions

```assembly
1: csrrsi	a0, mstatus, 8 # enable interrupts
2: wfi
3: csrrci	a0, mstatus, 8 # disable interrupts
```

Causes the core to sleep after `2` but immediately afterwards clears the interrupt enable from instruction `3`. Not sure if this is the right fix but it works for this problem

Before
irq-broken

After
irq-fixed

### DIFF
--- a/hw/ip/snitch/src/snitch.sv
+++ b/hw/ip/snitch/src/snitch.sv
@@ -722,14 +722,14 @@ module snitch import snitch_pkg::*; import riscv_instr::*; #(
         opb_select = None;
         rd_select = RdBypass;
         rd_bypass = csr_rvalue;
-        csr_en = 1'b1;
+        csr_en = valid_instr;
       end
       CSRRWI: begin
         opa_select = CSRImmmediate;
         opb_select = None;
         rd_select = RdBypass;
         rd_bypass = csr_rvalue;
-        csr_en = 1'b1;
+        csr_en = valid_instr;
       end
       CSRRS: begin  // Atomic Read and Set Bits in CSR
           alu_op = LOr;
@@ -737,7 +737,7 @@ module snitch import snitch_pkg::*; import riscv_instr::*; #(
           opb_select = CSR;
           rd_select = RdBypass;
           rd_bypass = csr_rvalue;
-          csr_en = 1'b1;
+          csr_en = valid_instr;
       end
       CSRRSI: begin
         // offload CSR enable to FP SS
@@ -747,7 +747,7 @@ module snitch import snitch_pkg::*; import riscv_instr::*; #(
           opb_select = CSR;
           rd_select = RdBypass;
           rd_bypass = csr_rvalue;
-          csr_en = 1'b1;
+          csr_en = valid_instr;
         end else begin
           write_rd = 1'b0;
           acc_qvalid_o = valid_instr;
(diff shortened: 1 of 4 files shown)


noah95:
There is a further bug regarding this:

```
1723 M 0x80000688 wfi    
                                           # !  interrupt hits
3965 M 0x8000068c csrrci  a0, mstatus, 8   # <- ERROR: insn after wfi in main progam is retired
3977 M 0x800027e0 addi    sp, sp, -80      # <- first instruction at mtvec
                                           # interrupt routine
4424 M 0x8000299c mret                     # <- return from interrupt
4429 M 0x8000068c csrrci  a0, mstatus, 8   # <- Correct issue of this instruction
```

`8000068c` is retired after the interrupt is triggered. This instruction should be issued AFTER the `mret` instruction. It is currently issued twice.

zarubaf:
Not sure I get the first issue, but the second one should probably be to disable any side-effects of instructions (CSR writes, register file writes) when we know that there is an interrupt associated with that instruction, right?

Can you elaborate a bit on the first issue a bit more? What would be the correct behavior?

noah95:
> Can you elaborate a bit on the first issue a bit more? What would be the correct behavior?

I expect the core to stall after the `wfi` instruction and jump to `mtvec` on an interrupt because interrupts are enabled in `mstatus`. The issue is that the core issues the instruction after the `wfi` which in this case disabled the interrupts. See the `interrupt` test in  #193 that sensitizes this bug.

> Not sure I get the first issue, but the second one should probably be to disable any side-effects of instructions (CSR writes, register file writes) when we know that there is an interrupt associated with that instruction, right?

What is an interrupt associated with an instruction? If you mean an interrupt pending, then yes. If we are stalling at a `wfi` and an interrupt is pending and interrupts in `mstatus` are enabled, the core should immediately jump to `mtvec` and not issue the instruction subsequent to the `wfi`

zarubaf:
Okay, I think the main problem is that interrupt that are associated with a `csr` write should not exhibit architectural side effects and the tracer should probably be adapted to mark the instruction as an interrupt.

> There is a further bug regarding this:
> 
> ```
> 1723 M 0x80000688 wfi    
>                                            # !  interrupt hits
> 3965 M 0x8000068c csrrci  a0, mstatus, 8   # <- ERROR: insn after wfi in main progam is retired
> 3977 M 0x800027e0 addi    sp, sp, -80      # <- first instruction at mtvec
>                                            # interrupt routine
> 4424 M 0x8000299c mret                     # <- return from interrupt
> 4429 M 0x8000068c csrrci  a0, mstatus, 8   # <- Correct issue of this instruction
> ```
> 
> `8000068c` is retired after the interrupt is triggered. This instruction should be issued AFTER the `mret` instruction. It is currently issued twice.

An interrupt always needs a corresponding instruction so, when the interrupt hits it gets associated with `3965 M 0x8000068c csrrci  a0, mstatus, 8`. The value in `mepc` is the current `pc`, so the `mret` jumps to the `epc`.

If the `3965 M 0x8000068c csrrci  a0, mstatus, 8` wouldn't exhibit side-effects the execution would be correct (except for the misleading tracer). That should also solve the `wfi` problem I guess.

zarubaf:
I've proposed another fix on this branch, once you are back I'd appreciate it if you could tell me whether this actually fixes something :-)

noah95:
cca3267  fixes the CSR access and 09bd868 adds an annotation on instructions where the `exception` signal was asserted. @zarubaf agree with the tracer or should we completely remove the instruction from the trace?

zarubaf (approved):
Perfect!

PR #195, opened 2021-08-09
ci: Fix editorconfig-checker version

noah95:
In `2.3.52` the binray name changed to `ec`. This fixes the version used to mitigate further CI lint problems


zarubaf (approved):
Awesome! Thanks!

PR #196, opened 2021-08-10
Multi Cluster runtime

fischeti:
This PR adds runtime support for multi-cluster systems like Occamy
### Changelog

- occamy: add new startup routine for occamy
- bootdata: add information about number of cluster, tcdm size & offset
- occamy: stack for each cluster is calculated based on `cluster_id`
- barrier: add software barrier for cluster and global synchronization
- occamy: trap manager core 0 in bootrom
- test: modify `printf_fmt.c` test for multicluster
- test: add new `multi-cluster.c` test
- barrier: move `barrier_reg_ptr` to team struct
- runtime: align runtime for `snitch_cluster` and `occamy`, currently named `start_occamy.[c|S]`
- occamy: auto-generate `bootdata.cc` from config file


hw/ip/test/src/tb_lib.hh
```diff
@@ -121,7 +121,12 @@ struct BootData {
     uint32_t core_count;
     uint32_t hartid_base;
     uint32_t tcdm_start;
-    uint32_t tcdm_end;
+    uint32_t tcdm_size;
+    uint32_t tcdm_offset;
+    uint32_t global_mem_start;
+    uint32_t global_mem_end;
+    uint32_t cluster_count;
+    uint32_t quadrant_count;
```
zarubaf, line 129: I think just to be future proof we want to call this `s1_*` as in a full Manticore system there are more than S1 quadrants.
```suggestion
    uint32_t s1_quadrant_count;
```

hw/ip/test/src/tb_lib.hh
```diff
@@ -121,7 +121,12 @@ struct BootData {
     uint32_t core_count;
     uint32_t hartid_base;
     uint32_t tcdm_start;
-    uint32_t tcdm_end;
+    uint32_t tcdm_size;
+    uint32_t tcdm_offset;
```
zarubaf, line 125: I was just wondering whether we want to make this an array and individual for each cluster e.g.,
```
uint32_t tcdm_addr [NR_CLUSTERS];
```

Because technically the TCDMs can be arbitrarily spaced. What do you think?

sw/snRuntime/src/start_occamy.c
```diff
@@ -0,0 +1,115 @@
+// Copyright 2020 ETH Zurich and University of Bologna.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+#include "snrt.h"
+#include "team.h"
+
```
zarubaf, line 6: I was wondering whether we maybe can make `start_occamy.[S|c]` the default that also works on the specialized case of the single cluster version, just to avoid bifurcation? Do you see troubles in going there?

I think the only thing that really needs to be adjusted is the `bootdata` struct that is generated from a script for the single-cluster version to indicate that there is just a single cluster, with a single S1 quadrant :).

sw/snRuntime/src/start_banshee.c
```diff
@@ -31,6 +31,14 @@ void _snrt_init_team(uint32_t cluster_core_id, uint32_t cluster_core_num,
     team->cluster_mem.start = spm_start;
     team->cluster_mem.end = spm_end;
 
+    // Allocate and initialize global barrier
+    // TODO: Atomics to global memory are currently
+    // not supported by banshee
```
zarubaf, line 36: Oh, okay, we might want to add this to Banshee, I'll open an issue for it :-)

zarubaf (approved):
LGTM! My biggest point is the discussion of whether we can merge the `occamy` runtime and the generic cluster runtime to avoid bifurcation. I think if we derive everything from the bootstruct we should be fairly good to go.

sw/snRuntime/src/start_banshee.c
fischeti: I am not 100% sure if banshee really doesn't support atomics to global memory. But I noticed that atomics to the global memory are definitely not coherent in the case of multi-cluster systems. So I assumed that it is not supported

SamuelRiedel: Atomics are supported in Banshee as well. What behavior are you observing? Can you provide me with a binary or test to debug this?

fischeti: Interestingly, I cannot reproduce the issue anymore :smile: 
I had the issue that the barrier on the main memory was treated as if it was private to each cluster i.e. the barrier did not increment to the number of cores globally but only to cores per cluster.

Currently, the atomics are working correctly when running banshee with `--trace`. However, it does not terminate when running without the trace option. Do you have an idea why this might be?

I am running the `printf_fmt` test from this branch:
`banshee --num-clusters 2 --num-cores 9 --no-opt-jit --no-opt-llvm --trace sw/build/snRuntime/test-snRuntime-printf_fmtint > log.dasm`

hw/ip/test/src/tb_lib.hh
fischeti: Makes sense

hw/ip/test/src/tb_lib.hh
fischeti: Could we auto-generate the `bootdata` somehow in that case? Otherwise this gets a bit tedious when switching between different configurations

sw/snRuntime/src/start_banshee.c
SamuelRiedel: Ok good to hear. The problem might have been fixed in PR https://github.com/pulp-platform/snitch/pull/188. The address map of Banshee was not aligned, which could have lead to the TCDM and global memory being swapped.

That it works with `--trace` sounds a bit like a race condition to me. I can have a closer look.

sw/snRuntime/src/start_occamy.c
fischeti: No I think that would make sense to align the two runtimes. Apart from the different `bootdata` struct they are pretty much the same.

How or where exactly is the `bootdata` generated for the `snitch_cluster`?

sw/snRuntime/src/start_banshee.c
fischeti: That sounds like the problem I had. I have re-built banshee with the new version at some point. This is would also explain why I can't reproduce the issue anymore

hw/ip/test/src/tb_lib.hh
zarubaf: Ah I see, we might want to make this a pointer so that we are not defining the size in the header and populate it correctly (using a script) in the `bootdata.cc`. Does that make sense?

I think for Occamy the `bootdata.cc` has been hard-coded so far, but for the single cluster version we should already be generating it from a script so we should extend that approach. I think that will become more and more useful, especially if we start to describe more complex systems.

sw/snRuntime/src/start_occamy.c
zarubaf: For the cluster it is generated here: https://github.com/pulp-platform/snitch/blob/078a14ab715a759a6146609cea3fba53a3dcf4c2/util/clustergen/cluster.py#L357-L360

For `occamy` we probably want to do something very similar (same template) but [here](https://github.com/pulp-platform/snitch/blob/078a14ab715a759a6146609cea3fba53a3dcf4c2/util/occamygen.py)

hw/ip/test/src/tb_lib.hh
fischeti: > Ah I see, we might want to make this a pointer so that we are not defining the size in the header and populate it correctly (using a script) in the `bootdata.cc`. Does that make sense?

Not quite, this `bootdata` struct is defined in three places `tb_lib.hh`, `start_occamy.c` and `bootdata.cc`. As far as I understand, those need to match

sw/snRuntime/src/start_occamy.c
fischeti: I have now aligned the two runtimes and the `bootdata` is generated automatically based on the configuration file also for occamy.

hw/ip/test/src/tb_lib.hh
zarubaf: As discussed in the daily today. @noah95 will tackle that in the next weeks. We are good for now.

zarubaf (approved):
LGTM

hw/system/snitch_cluster/cfg/cluster.default.hjson
```diff
@@ -62,10 +69,10 @@
         ]
     },
     dram: {
-        // 0x8000_0000
-        address: 2147483648,
-        // 0x8000_0000
-        length: 2147483648
+        // 0x9000_0000
```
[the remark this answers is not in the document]
fischeti: Kind of, `snrt_global_memory()` is initialized with those values. If it would start at `0x8000_0000` you would potentially overwrite the `.text` section, if `snrt_global_memory()` is used for allocating data.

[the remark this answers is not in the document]
zarubaf: Okay, but you could add the `0x1000_0000` offset in `snrt_global_memory()` as well, right? Seems kind of cheating to artificially reduce the memory in the system description (which should be a true and holistic view of the system).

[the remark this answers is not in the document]
fischeti: True, I will change this accordingly

sw/snRuntime/src/start_banshee.c
```diff
@@ -31,6 +31,12 @@ void _snrt_init_team(uint32_t cluster_core_id, uint32_t cluster_core_num,
     team->cluster_mem.start = spm_start;
     team->cluster_mem.end = spm_end;
 
+    // Allocate and initialize global barrier
+    team->global_barrier = (void *)team->global_mem.start;
```
noah95, line 35: Same problem here

noah95 (changes requested):

sw/snRuntime/src/start_cluster.c
```diff
@@ -33,7 +37,7 @@ struct putc_buffer_header {
 static volatile struct putc_buffer {
     struct putc_buffer_header hdr;
     char data[PUTC_BUFFER_LEN];
-} *const putc_buffer = (void *)&_edram;
+} *const putc_buffer = (void *)&_end;
```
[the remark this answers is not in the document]
noah95: Moved back to `_edram` in d115508

sw/snRuntime/src/start_cluster.c
```diff
@@ -42,25 +46,37 @@ void _snrt_init_team(uint32_t cluster_core_id, uint32_t cluster_core_num,
...
     team->cluster_mem.start = spm_start;
     team->cluster_mem.end = spm_end;
+    team->barrier_reg_ptr = spm_start + bootdata->tcdm_size + 0x30;
+
+    // Initialize cluster barrier
+    team->cluster_barrier.barrier = 0;
+    team->cluster_barrier.barrier_iteration = 0;
+
+    // Allocate and initialize global barrier
+    team->global_barrier = (void *)team->global_mem.start;
```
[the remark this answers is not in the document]
noah95: Move global barrier to `.dram` section in d115508

sw/snRuntime/src/team.c
```diff
@@ -110,3 +112,44 @@ void snrt_bcast_recv(void *data, size_t len) {
     snrt_memcpy(data, mbox->ptr, len);
     snrt_barrier();
 }
+
+/// Synchronize cores in a cluster with a software barrier
+void snrt_cluster_barrier() {
```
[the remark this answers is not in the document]
noah95: Renamed cluster barriers to reflect hw/sw barrier

noah95 (approved):
@fischeti if you agree with the changes in d115508 we can merge :+1:

fischeti:
> @fischeti if you agree with the changes in [d115508](https://github.com/pulp-platform/snitch/commit/d115508c02a66508b8f417f7e380809a56a7788c) we can merge +1

LGTM! Let's merge it

PR #199, opened 2021-08-13
fpss: Tag SSR writebacks to prevent CSR toggle glitches

paulsc96:
When the SSR redirect CSR (`0x7C0`) is toggled, inflight writes to SSR register addresses may not be written back correctly:

* On enable: inflight writes to regular registers will be directed to SSRs
* On disable: inflight writes to SSRs will be directed to regular registers

This can cause frustrating glitches, especially because SSRs which overrun their jobs may block the FP pipeline indefinitely.

This PR determines at decode whether an instruction writes back to an SSR instead and propagates this information through the FPU with an additional tag bit.

This should be very lightweight in terms of additional logic and ensure "precise" SSR toggling, i.e. that SSRs are accessed exactly as defined by the program flow regardless of execution timing.


zarubaf (approved):
Cool, that is definitely the cleaner version! Potentially saves a lot of pain :-D Thanks

zarubaf:
Thanks, Paul!

PR #200, opened 2021-08-16
occamy: Provide wide to narrow connection

zarubaf:
As discussed in #192 the missing wide to narrow connection makes it impossible for the Snitch clusters to fetch from the bootrom. This PR fixes that issue and should probably go before #192.


niwis (approved):
Looks good to me :-)

PR #201, opened 2021-08-16
lint: Fix missing signals and tb lint problems

zarubaf:
Seems that signal was undriven in the case of de-activated SSRs.


zarubaf:
Re-based on current master. Not sure whether you already included it @paulsc96 - so please merge or close if this is still relevant.

paulsc96 (approved):
LGTM; I have not fixed this yet; doing it in its own PR is cleaner.

PR #202, opened 2021-08-16
Run CI on pull requests

noah95:
I propose that we run the CI on pull requests so that PRs from external repositories also run the CI (e.g. #174 )


zarubaf:
LGTM!

PR #206, opened 2021-08-18
Move SPL to SPM

niwis:
Move the SPL into SPM and boot from there. Furthermore, fix a couple of issues that were encountered on the way. Most importantly:

* Update the SPL contained in the bootrom according to https://github.com/pulp-platform/u-boot/pull/4 and change the ZSBL's SPL destination address to the SPM
* Cut a combinatorial loop through the SPM's atomic adapter
* Update Ariane's PMAs


zarubaf (approved):
LGTM! Thanks

PR #209, opened 2021-08-23
Read-only Cache

SamuelRiedel:
This PR adds a read-only cache that can be inserted into the AXI hierarchy. It features the following:

- AXI master and slave port to be inserted directly into the AXI bus
- Address map port to configure which address regions are cached and which will be forwarded upstream
- The cache line has to be larger or equal to the AXI data width
- It supports bursts across multiple cache lines, but they have to use the full AXI data width per beat

## To discuss

- [x] Rename `const cache` to `read-only cache`
- [x] How to handle changes in the `snitch_icache` modules
- [x] The mux and demux around the cache do not guarantee that a request to a cacheable region and one forwarded upstream are not reordered. This violates the AXI spec but is quite expensive to fix.

## Todo

- [x] Finish clean-up
- [x] Check the max frequency in the real system


hw/ip/snitch_icache/src/snitch_icache_handler.sv
```diff
@@ -142,6 +142,29 @@ module snitch_icache_handler #(
...
+        end
+    end
+
+    always_ff @(posedge clk_i or negedge rst_ni) begin
+       if(~rst_ni) begin
+           miss_in_flight_q <= '0;
+       end else begin
+           miss_in_flight_q <= miss_in_flight_d;
+       end
+    end
```
zarubaf, line 166: Would you mind making this the apropriate `FF` macro?

zarubaf (approved):
That looks pretty good to me - thanks *a lot* for this contribution. The only point of concern is the bypass of sub-bus accesses. We should probably discuss this (and maybe assess how much more costly this would be to support, in terms of HW and time to implement).

Also really cool that you found some bugs in the instruction cache!

zarubaf:
> * [ ]  Rename `const cache` to `read-only cache`

I am in favour of re-naming it now before it is used everywhere :) it will just become more effort in my experience

> * [ ]  How to handle changes in the `snitch_icache` modules

I think the changes don't really impact the icache too much so we can just use them everywhere, whats your take on it?

> * [ ]  The mux and demux around the cache do not guarantee that a request to a cacheable region and one forwarded upstream are not reordered. This violates the AXI spec but is quite expensive to fix.

Indeed, I am not overly concerned about that since we are dealing with reads to idempotent memory regions. Read re-ordering should be okay. Writes are in order since they will always bypass. And the `demux` will take care that we are not doing any protocol violation on the incoming port. Do you agree?

zarubaf:
@SamuelRiedel Are we good to merge? I can see that all boxes are ticked :-)

SamuelRiedel:
From my side, we are good to go. I was waiting for your opinion on the ID queue issue we discussed in the last meeting. Updating it would improve performance, but we can also do this in a separate PR.

zarubaf:
I am sorry, I've totally missed the conversation. I will patch the ID queue locally and merge it upstream so that we have all the performance here. I guess just enabling the "parallel" ID Queue feature is all that is need as a change in the RO cache, right?

SamuelRiedel:
Yes exactly. You can also let me know once the new ID queue is in Occamy and I can update the cache. But it is really just setting the parameter to 1.

PR #211, opened 2021-08-24
vendor: Fix reqrsp_interface dependency

fischeti:
Fixes issue where bender fails to find correct path for the `reqrsp_interface` ip.


zarubaf:
Thanks @fischeti! Looks good, I think we should merge even though the Banshee build seems to fail conistenly.

PR #212, opened 2021-08-25
occamy: Add HBM

zarubaf:
There are a couple of fixes necessary here to integrate the HBM. Nothing major, a couple of interrupts more plus the usual APB configuration bus.

### DIFF
--- a/util/occamygen.py
+++ b/util/occamygen.py
@@ -184,6 +184,9 @@ def main():
     am_hbi_ctl = am.new_leaf("hbi_ctl", 0x10000,
                              0x07000000).attach_to(am_soc_regbus_periph_xbar)
 
+    am_hbm_cfg = am.new_leaf("hbm_cfg", 0x10000,
+                             0x08000000).attach_to(am_soc_regbus_periph_xbar)
+
     am_plic = am.new_leaf("plic", 0x4000000,
                           0x0C000000).attach_to(am_soc_regbus_periph_xbar)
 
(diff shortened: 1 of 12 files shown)


niwis (commented):
Looks very good to me in general. Just one thought: Maybe we should tie the new APB response port and the new interrupt inputs in `occamy_xilinx.sv.tpl`? If you like, I can also quickly do that.

util/occamygen.py
```diff
@@ -391,11 +401,13 @@ def main():
...
         "soc_regbus_periph_xbar": soc_regbus_periph_xbar,
         "apb_hbi_ctl": apb_hbi_ctl,
+        "apb_hbm_cfg": apb_hbm_cfg,
         "cfg": occamy.cfg,
         "cores": nr_s1_quadrants * nr_s1_clusters * nr_cluster_cores + 1,
         "nr_s1_quadrants": nr_s1_quadrants,
         "nr_s1_clusters": nr_s1_clusters,
-        "nr_cluster_cores": nr_cluster_cores
+        "nr_cluster_cores": nr_cluster_cores,
+        "hbm_channel_size": HBM_CHANNEL_SIZE
```
niwis, line 410: I cannot find `hbm_channel_size` being used anywhere — or am I wrong?

zarubaf: Yeah indeed, it is needed in the private repo `snitch-pd` :-)

niwis: 👍

PR #213, opened 2021-08-25
Improve AXI Isolation

zarubaf:
An isolated port could mean hanging the entire system. This commit
improves that by optionally terminating AXI transaction in case the
adapter is isolated.


niwis:
FPGA build is running

PR #217, opened 2021-08-26
ssr: Multi-SSR index intersection system and stream-controlled FREP

paulsc96:
Adds the internally-discussed intersection system between SSRs. Among other possible uses, it enables basic sparse-sparse linear algebra (addition, dot products, element-wise multiplication) between sparse vectors and therefore between higher sparse tensors.

In a new mode, two indirection-capable _master_ SSRs may compare their ordered index streams in a shared _intersector_ to form a streaming intersection or union of indices. For each resulting index, an address is emitted from a linear sequence; in case of an index union, the SSR lacking an index emits a dummy `'0` element. The intersected indices may optionally be redirected to a third _slave_ SSR which uses its index port to write out the resulting index stream. For each such index, it again emits an address,  which may be used to store data associated with the index.

Since the number of address emissions during intersection is unknown, but common to both _master_ and  _slave_ SSRs, a new FREP mode (`frep.s`) was defined: this mode is activated by re-purposing the uppermost bit of the `max_inst` field, which was well over-provisioned at 12 bits. It is controlled by the shared intersector through a generic interface, handshaking once for each emission and asserting a done signal to terminate the loop.

Like the existing SSR system, all features are fully parameterizable. Likewise, all existing SSR functionality is maintained. The functionality was tested in a standalone testbench, maintaining existing tests to prevent regression, and in external software tests. 

A default parameterization of the system is added to the Occamy configuration.

In summary, this PR adds:

* The Multi-SSR index intersection system with intersector, master SSRs, and slave SSRs
* A stream-controlled FREP loop (`frep.s`)
* A standalone SSR streamer testbench verifying the intersection and union features
* Various small improvements to existing RTL
* Intersecting SSRs to Occamy

It does _not_ yet add (left to future PRs):

* Banshee support for intersection
* Runtime / software support


zarubaf (approved):
Thanks, @paulsc96 - quite an extensive change but looking over it, it looks good to me!

PR #221, opened 2021-08-31
runtime: Cancel tests after timeout

noah95:
This change causes stalling unit tests to fail after 3/30 minutes (banshee/RTL).


noah95:
Why would [some tests](https://github.com/pulp-platform/snitch/runs/3472741444?check_suite_focus=true) suddenly stop with `Error: The operation was canceled.`?

zarubaf:
That is a good question. Maybe because GitHub sees that there is already a CI test failing and therefore killing the other one to free the resources?

zarubaf (approved):
Elegant fix in CMake. Thanks

noah95:
> That is a good question. Maybe because GitHub sees that there is already a CI test failing and therefore killing the other one to free the resources?

Time to fix the failing tests then ;)

PR #222, opened 2021-08-31
solder: Make full use of target ID space in IWC

stmach:
Use all IDs in the remapped space instead of a static 4. I'm not sure what the impact on HW complexity is, so please chip in!


niwis (approved):
Looks good, `axi_remap` scales linearly with `MaxUniqIds`.

zarubaf:
Generally looks good to me. Let's keep this in mind when we see that there is a bottleneck during physical design. CC/ @yichao-zh

PR #225, opened 2021-09-02
Increase Banshee timeout to 360s

zarubaf:
I think the 180s are a bit too optimistic for one of the tests. Let's double it, still better than 5h :-)


noah95 (approved):

zarubaf:
> Agree 👍 Did you manually catch the trailing white-space or was that your editor and if so, which and configured how?

That was my editor :) VS Code using the `editorconfig` plugin which should pick up the `.editorconfig` in the root of this project. But I think I also manually configured it to trim all whitespaces.

PR #227, opened 2021-09-02
banshee: Set mcause on interrupt even if mstatus.mie is 0

noah95:
An interrupt without global interrupt enable would not set the `mcause` CSR correctly. This fixes it and changes the interrupt test to sensitize this bug

### DIFF
--- a/sw/banshee/src/tran.rs
+++ b/sw/banshee/src/tran.rs
@@ -5631,18 +5631,24 @@ impl<'a> InstructionTranslator<'a> {
             LLVMConstInt(LLVMInt32Type(), 0, 0),
             NONAME,
         );
-        let is_irq = LLVMBuildAnd(
-            self.builder,
-            is_mstatus_mie,
-            is_any_interrupt_pending,
-            NONAME,
+
+        // Create BB for any pending interrupt
+        let bb_any_irq = LLVMCreateBasicBlockInContext(self.section.engine.context, NONAME);
+        LLVMInsertExistingBasicBlockAfterInsertBlock(self.builder, bb_any_irq);
+        LLVMBuildCondBr(self.builder, is_any_interrupt_pending, bb_any_irq, bb_noirq);
+        LLVMPositionBuilderAtEnd(self.builder, bb_any_irq);
+
+        // Set mcause register
+        // TODO: Get correct cause, currently only settings machine software interrupt
+        self.write_csr_silent(
+            riscv::Csr::Mcause as u32,
+            LLVMConstInt(LLVMInt32Type(), (1 << 31) | 3, 0),
         );
-        // Create BB for IRQ
+
+        // Create BB for IRQ and mstatus.mie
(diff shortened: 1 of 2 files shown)


zarubaf (approved):
Nice catch!

SamuelRiedel (commented):
LGTM.

noah95:
This does not coincide with the current hardware implementation: If snitch exits `wfi` due to an interrupt while `ie_q[M]=0` (`mstatus.mie=1`) which I use in the second part of the interrupt test (wake from interrupt but don't jump to `mtvec`), the `mcasue` register is not written by HW:

https://github.com/pulp-platform/snitch/blob/ed09f0e12fefb52567b739b2883a07c26cb024aa/hw/ip/snitch/src/snitch.sv#L2271-L2279

Since `interrupt` is only asserted if `ie_q[M]=1`

https://github.com/pulp-platform/snitch/blob/ed09f0e12fefb52567b739b2883a07c26cb024aa/hw/ip/snitch/src/snitch.sv#L2019-L2021

IMO the [privilege spec](https://riscv.org/wp-content/uploads/2017/05/riscv-privileged-v1.10.pdf) does not distinguish between trap taken from "normal" and "wfi" state, thus the `mcasue` register should be written in this case.

@zarubaf What's your take on this?

zarubaf:
Okay, not 100% sure I get the point:

> IMO the privilege spec does not distinguish between trap taken from "normal" and "wfi" state, thus the mcasue register should be written in this case.

Agree, especially because there is no `wfi` state defined in the spec.

> This does not coincide with the current hardware implementation: If snitch exits `wfi` due to an interrupt while `ie_q[M]=0` (`mstatus.mie=1`) which I use in the second part of the interrupt test (wake from interrupt but don't jump to `mtvec`), the `mcasue` register is not written by HW

That seems correct to me. If the core does not jump to `mtvec` then the interrupt is not considered taken hence, no update of `mcause`. That the core exits `wfi` is a special property of that instruction so that forward progress can be made.

Did this answer your question?

noah95:
> > This does not coincide with the current hardware implementation: If snitch exits `wfi` due to an interrupt while `ie_q[M]=0` (`mstatus.mie=1`) which I use in the second part of the interrupt test (wake from interrupt but don't jump to `mtvec`), the `mcasue` register is not written by HW
> 
> That seems correct to me. If the core does not jump to `mtvec` then the interrupt is not considered taken hence, no update of `mcause`. That the core exits `wfi` is a special property of that instruction so that forward progress can be made.

So there is no way in telling what caused the hart to exit from `wfi` and `mcasue` is untouched? IF that is the case, we have to close this PR since the HW doesn't touch `mcause` if exited from `wfi` with interrupts disabled.

zarubaf:
That’s right. `wfi` could also be legally implemented as a `nop` according to the spec.

noah95:
> That’s right. `wfi` could also be legally implemented as a `nop` according to the spec.

Correct, thanks. This settles this discussion :) Closing

PR #229, opened 2021-09-02
banshee: Change amoadd to a wrapping addition

noah95:
This fixes any `attempt to add with overflow` panicks when using `amoadd`. I asusme wrapping add is the correct behaviour?


SamuelRiedel (approved):
As far as I know, this is not explicitly specified in the RISC-V ISA manual, but this is what we do in HW and is most intuitive. Thanks for fixing this :)

PR #230, opened 2021-09-02
docker: Change LLVM toolchain URL

noah95:
The `snitch-llvm` repository moved to `llvm-project` which causes the container build to fail


zarubaf:
I thought GitHub redirects... Anyway LGTM

noah95:
> I thought GitHub redirects... Anyway LGTM

It might not redirect the API requests

PR #231, opened 2021-09-03
cva6: Make L1D write-back

niwis:
Exchange Ariane's write-through L1 data cache with a write-back.

- [x] Exchange caches
- [x] Make the WB's AXI port width configurable
- [x] Mark the WB's AXI requests as modifiable
- [x] Add instruction fence after zero-stage bootloader
- [x] Compile for FPGA
- [x] Boot Linux
- [x] Pray
- [x] Update vendor patch


zarubaf:
> * [ ]  Pray

I'll being praying for you :-D

niwis:
So Linux boots now! The flash loading sequences (U-Boot SPL and U-Boot) seem significantly slower though. But this is just a first impression, I would need to profile this for more details.

zarubaf:
Oh wow, that is great! What's your take? Should we merge? I guess we have seen this in the past with Linux as well, right - as the WB cache is actually writing-back and invalidating on a `fence` that takes longer.

niwis:
As long as it's just the `fence` I think we should be fine; in my opinion, boot speed isn't the primary goal. I'm just a bit worried that there might be something else going on (e.g. slow evictions), as I don't see why there needs to be more than one `fence` after loading from flash, and that wouldn't make the difference. But perhaps U-Boot is as generous with `fence`s as Linux :-)

zarubaf (approved):
LGTM, we should probably also get a couple of patches applied to upstream CVA6 (the modifiable and port parameterization).

Let's discuss in tomorrow's weekly about the slowness. Then we can merge.

zarubaf:
Thanks, @niwis, according to the discussions today let's merge.

PR #232, opened 2021-09-03
occamy_pkg: Derive number of clusters from cfg

niwis:
Derive the number of clusters from `occamy_cfg.hjson`. This fixes port
mismatches for numbers of clusters different from the default value
(`4`)


zarubaf (approved):
Smart!

PR #233, opened 2021-09-04
occamy: Integrate RO cache

zarubaf:


hw/system/occamy/src/occamy_soc_ctrl/occamy_soc_reg.hjson
```diff
@@ -164,6 +164,743 @@
...
+        desc: "Flush read-only cache.",
+        swaccess: "rw"
+        hwaccess: "hrw"
+        count: "NumS1Quadrants"
+        cname: "ro_flush"
+        fields: [
+          { bits: "0:0"
+            resval: "0"
+            name: "FLUSH"
+            desc: "Flush (inv RO cache of S1 quadrant."
```
SamuelRiedel, line 195: Nitpick: Missing closing paranthesis.

hw/system/occamy/src/occamy_soc_ctrl/occamy_soc_reg.hjson.tpl
```diff
@@ -0,0 +1,250 @@
...
+        desc: "Flush read-only cache.",
+        swaccess: "rw"
+        hwaccess: "hrw"
+        count: "NumS1Quadrants"
+        cname: "ro_flush"
+        fields: [
+          { bits: "0:0"
+            resval: "0"
+            name: "FLUSH"
+            desc: "Flush (inv RO cache of S1 quadrant."
```
SamuelRiedel, line 195: I guess the missing closing parenthesis comes from here.

hw/system/occamy/src/occamy_top.sv
```diff
@@ -710,6 +710,22 @@ SOC_REGBUS_PERIPH_XBAR_NUM_OUTPUTS
...
+  assign end_addr_0[0] = {
+    soc_ctrl_out.ro_end_addr_high_0_quadrant_0.q, soc_ctrl_out.ro_end_addr_low_0_quadrant_0.q
+  };
+  assign start_addr_0[1] = {
+    soc_ctrl_out.ro_start_addr_high_1_quadrant_0.q, soc_ctrl_out.ro_start_addr_low_1_quadrant_0.q
+  };
+  assign end_addr_0[1] = {
+    soc_ctrl_out.ro_end_addr_high_1_quadrant_0.q, soc_ctrl_out.ro_end_addr_low_1_quadrant_0.q
+  };
+  assign soc_ctrl_in.ro_cache_flush[0].de = soc_ctrl_out.ro_cache_flush[0].q & soc_ctrl_in.ro_cache_flush[0].d;
```
SamuelRiedel, line 727: What is this used for?

This is most likely not related, but I just noticed that trying to flush the lookup module, while it is already in the flush state, might lead to a deadlock on the flushing protocol. If you don't get the ready [immediately](https://github.com/pulp-platform/snitch/blob/d6d5d545c727cc98dfb55c34aa85c978c515e293/hw/ip/snitch_icache/src/snitch_icache_lookup.sv#L92), you will [never get it](https://github.com/pulp-platform/snitch/blob/d6d5d545c727cc98dfb55c34aa85c978c515e293/hw/ip/snitch_icache/src/snitch_icache_lookup.sv#L97). This is something to fix, or to keep in mind when setting the `flush_valid` signal.

SamuelRiedel (approved):
LGTM. I have a minor point to discuss about the flushing handshake.

hw/system/occamy/src/occamy_top.sv
zarubaf: Hmm, that should simulate a handshake on the flush logic. And make it possible to observe the current state of the flush (although I agree, as the ready comes in the first cycle that point is moot). But I don't see the deadlock (assuming the `flush_ready`, `flush_valid` has AXI-style handshake semantic).

hw/system/occamy/src/occamy_soc_ctrl/occamy_soc_reg.hjson.tpl
zarubaf: Yes, thanks!

hw/system/occamy/src/occamy_top.sv
SamuelRiedel: I see. The problem I noticed is not with the code here, but with the implementation of the handshaking in the `lookup` module, which is not AXI-style in that sense. It will not start the flush while the `valid` signal is applied, but it will only give the ready signal if it is not in the flush state. So applying a `valid` while the `lookup` module is already flushing will lead to it never responding with a `ready`.

zarubaf: Ooops, indeed. I think we should definitely fix that probably (in the flush logic). Would it be possible to take care of that in a separate PR?

fischeti (approved):
LGTM!

hw/system/occamy/src/occamy_top.sv
SamuelRiedel: Yes, I can open a new PR to handle this.

niwis (approved):
Same here :-)

PR #235, opened 2021-09-06
banshee: Disable IR and JIT optimizations per default

noah95:
I propose to disable module and jit optimization since in most use-cases optimization takes longer than actual runtime.


SamuelRiedel (approved):
This is fine for me. :+1:

fischeti (approved):

PR #237, opened 2021-09-06
runtime: OpenMP support

noah95:
Okay, another big one:

### runtime
- **event unit** (`eu.c`): All but hart 0 of a cluster enter the event loop. A function with arguments can be pushed to the event queue (`eu_dispatch_push`) by the master hart and then the specified number of threads execute if on `eu_run_empty`. Synchronization is done through the wakeup register and atomics in TCDM. A `CLINT` implementation is also available but is less performant
- **data mover** (`dm.c`): Small event queue specialized for the DM core. Transactions can be queued with `dm_memcpy[2d]_async` and synchronized with `dm_wait`
- **OpenMP** (`omp.c`,`kmp.c`): Basic OpenMP support for static and dynamic scheduling. The `OMPSTATIC_NUMTHREADS` cmake variable can be set to the number of worker cores for a static implementation of OpenMP that is quicker but doesn't support dynamic scheduling
- **test** for `event_unit`, `datamover` and `openmp` that also serve as a usage example
- **snrt** `snrt_memset` utility
- **wakeup** Wakeup register added as peripheral in the `snrt_peripherals` struct. Writes can be done through `snrt_wakeup`
- OpenMP bootstrap macros
- Simple L3 memory allocator
- **interrupt**: clint pointer to TLS for faster access
- **interrupt**: Polling method that reduces the stress on the clint mutex
- `core_idx` is now cached in TLS so that is not computed on each call to `snrt_cluster_core_idx`

### banshee
- Remove base_hartid offset in wakeup register to make consistent with RTL (#246)
- Default to 3 SSR DM

### util
- **annotate** Make `annotate.py` RTL-trace friendly
- **wave** Extended `all_cores.do` wave script with SSR
- **tracevis** Extended with look-ahead so that long load instructions are displayed as such. Works only for RTL traces of course

### ci
- Test dynamic and static openmp implementation

## ToDo
- [x] Put eu/omp data into cluster TCDM if possible
- [x] interrupt: Single mutex lock and check/clear interrupt if interrupt present in clint
- [ ] Get overhead numbers for launch (omp parallel, dm memcpy, ?)
- [ ] make OMPSTATIC_NUMTHREADS=8 the default?


noah95:
Intra-cluster synchronization between worker and data mover core is not guaranteed to work with wakeup register (race conditions), hence the default is now to use the CLINT. Will port this to the cluster-local CLINT once implemented (#248).

sw/snRuntime/src/start_cluster.c
```diff
@@ -55,25 +56,21 @@ void _snrt_init_team(uint32_t cluster_core_id, uint32_t cluster_core_num,
...
-
-    // Allocate memory for a cluster mailbox.
-    team->cluster_mem.end -= sizeof(struct snrt_mailbox);
-    team->cluster_mailbox = (void *)team->cluster_mem.end;
-
+    // TLS caches of frequently used data
     _snrt_team_current = &team->base;
+    _snrt_core_idx =
+        (snrt_hartid() - _snrt_team_current->root->cluster_core_base_hartid) %
+        _snrt_team_current->root->cluster_core_num;
```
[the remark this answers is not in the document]
noah95: Yes absolutely! There is a clear trade-off here between cycles overhead and memory usage. I noticed this when inspecting with `tracevis.py` and moved the heavy-hitters to TLS. If you find any other data that has this characteristic, let me know

util/trace/tracevis.py
```diff
@@ -137,24 +146,27 @@ def parse_line(line, hartid):
...
+    #     if match:
+    #         (time, cyc, priv, cmt) = tuple(
+    #             [match.group(i+1).strip() for i in range(re_acc_line.groups)])
+    #         # use time,cyc from last line
+    #         time, cyc = last_time, last_cyc
+    #         args = pc = ''
+    #         instr = 'acc'
+    #     else:
+    #         print(f'no match in "{line}"')
+    #         return 1
```
zarubaf, line 165: Any reason to keep this?

zarubaf (approved):
Great work @noah95!

.github/workflows/ci.yml
```diff
@@ -46,6 +46,9 @@ jobs:
         toolchain:
         - gcc
         - llvm
+        cmake_args:
+        - ''
```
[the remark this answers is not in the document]
noah95: To run the tests with the default value of `OMPSTATIC_NUMTHREADS`

fischeti (approved):

util/trace/tracevis.py
noah95: Nope

sw/banshee/src/engine.rs
```diff
@@ -696,8 +697,10 @@ impl<'a, 'b> Cpu<'a, 'b> {
                     }
                 } else if (value as usize) < self.wake_up.len() {
                     self.num_sleep.fetch_sub(1, Ordering::Relaxed);
-                    self.wake_up[value as usize - self.engine.base_hartid]
-                        .fetch_max(self.state.cycle + 1, Ordering::Release);
+                    self.wake_up[value as usize].fetch_max(self.state.cycle + 1, Ordering::Release);
```
SamuelRiedel, line 700: This changes how the `wake_up` vector is interpreted. Shouldn't this change also be reflected in the `binary_wfi` and in the clint functions?

SamuelRiedel (approved):
LGTM.

paulsc96 (approved):
From the Banshee changes alone, looks good to me :) Except for the tidbit @SamuelRiedel already pointed out.

sw/banshee/src/engine.rs
noah95: `binary_wfi` already interprets `wake_up` as "offset-free" (entry `0` being the first hart, regardless of `base_hartid`). This is a bug that was not caught by the banshee `wfi` test. Do you agree?

https://github.com/pulp-platform/snitch/blob/fcc67a79f2d4edb31c1ddd17167fb75251d04951/sw/banshee/src/engine.rs#L914-L915

SamuelRiedel: I think this depends on how you handle it in software. If your `base_hartid` is 10, and you want to wake up the core with `hartid` 10, do you write the value `10` or `0` to the wake_up register? My assumption was that you write the `hartid` (10), so you have to remove the offset in Banshee. In your updated version, you'd have to write a zero, right? Meaning the software already removes the offset.

noah95: Correct, the offset is removed in software because that is also how it is in RTL (at least in Snitch, i don't know about mempool). RTL wires bit 0 of the wakeup register to the first hart in the cluster, regardless of the base hartid

SamuelRiedel: Perfect. In this case, everything is good to go from my side. In MemPool, we don‘t have an offset.

PR #238, opened 2021-09-06
occamy: Correct parameterization of DM core

paulsc96:
Removes floating-point features from the datamover core in each Snitch Cluster of Occamy to agree with the default parameterization.


niwis (approved):
LGTM

zarubaf (approved):
Thanks, @paulsc96 for catching this!

PR #240, opened 2021-09-07
ssr: VCS elaboration fixes

paulsc96:
This fixes some errors and warnings reported by VCS due to SSRs during elaboration of Occamy.


zarubaf (approved):
Classics, stumbled over that a couple of times as well. LGTM

PR #241, opened 2021-09-07
occamy: Update generator to include read-only cache memories

paulsc96:
`occamygen` was outdated, resulting in read-only cache SRAMs not being included in the generated `memories.json`. This PR fixes this.


zarubaf (approved):
Perfect, thanks!

PR #243, opened 2021-09-07
soc-regs: Stagger ro-cache reset regions to avoid illegal address rules

paulsc96:
Currently, the default read-only cache address regions cause a fatal assertion failure on simulation. This assertion is unfortunately worded, but valid, as the address decoder used is not designed for address ranges of size 0 or smaller and spews further assertion failures if the former is circumvented.

This MR sets sensible default values for the SoC registers setting the address rules: each rule is set to a 32-bit block continuously addressed from 0 on reset. this should not be an issue, as the cache is bypassed by default.

This could be improved upon with configurable default address ranges, but this might be overkill and needs further configuration validation.


zarubaf (approved):
While I do think this is the wrong fix for the problem, let’s merge this for now as I don’t see any immediate drawbacks and it allows us to make forward progress.

PR #244, opened 2021-09-08
Fix cache flush signal

SamuelRiedel:
This PR fixes a few issues related to the cache flushing logic:

- Each core will only flush its own L0 cache.
- The flush ready signal is always asserted since the cache is always ready to flush.
- The lookup module will accept new flush requests while it is flushing, but it will not restart the flush procedure.

Additionally, it adds a test to check if one of the cores deadlocks while flushing the cache simultaneously with the other cores.


SamuelRiedel:
@zarubaf: The container test failing is expected when adding a new instruction to Banshee, right?

zarubaf (approved):
Great, thanks a lot @SamuelRiedel! Yes, the test is supposed to fail. Let's merge!

PR #249, opened 2021-09-15
ci: Remove linting of history

zarubaf:
As the workflow here is squash and merge we do not need to have a look at the commit messages to guarantee a linear history.


noah95:
Agree. We could instead check the PR title which becomes the squashed commit for compliance?

zarubaf:
Yes, fully agree. That would be awesome. Seems to be quite a rabbit hole though so I decided to postpone this and we can just merge now imho. I will open an issue :)

PR #250, opened 2021-09-15
occamy: Fix const cache mismatches

zarubaf:
Okay, here is my untested try to fix the mismatches. I've also added an assertion so that this doesn't happen again. Probably also wise to add the check to the other cluster ports.


paulsc96:
Thanks! I will do some checks to make sure this resolves our issues.

paulsc96 (approved):
It seems to work in simulation :) I can de-isolate an individual quadrant and gate the fetching of Snitch clusters accordingly. Good to go for me :+1:

PR #252, opened 2021-09-15
occamy: Update CVA6 memories to write-back migration

paulsc96:
The CVA6 memories declared in `occamygen.py` still reflected the write-through data cache configuration; now, they declare the correct write-back memories.


zarubaf (approved):
Thanks!

PR #253, opened 2021-09-15
occamy: Isolate all S1 quadrants by default

paulsc96:
This ensures that all S1 quadrants have all of their master and slave ports isolated on reset by changing the respective SoC registers' reset values.


zarubaf (approved):
Thanks for catching this!

PR #254, opened 2021-09-15
ro-cache: Fix Bender manifest dependencies

paulsc96:
The read-only cache's Bender manifest contains wrong paths and is missing a dependency on `common_cells`. This currently causes no issues, but we could eventually trip over this as we did with `reqrsp_interface`.


zarubaf (approved):
Right, absolutely true! Thanks

Issue #247, opened 2021-09-13
Add DMA performance metrics to perf counters


zarubaf:

zarubaf:
@fischeti as you are looking into the performance counters at the moment would you mind also tackling that issue? Thanks!

fischeti:
Sure 👍

fischeti:
https://github.com/pulp-platform/snitch/blob/ca929ad402878b64ab3746d86fb25ae6d245930f/hw/ip/snitch_dma/src/axi_dma_pkg.sv#L13-L23

Those are quite a lot of and big counters. I guess the idea is not to synthesize them but connect them to the performance counter in an event-based style, right?

For example as in : 
https://github.com/pulp-platform/snitch/blob/be5622755f31d71548a22508f98451e03f7eb125/hw/ip/snitch_cluster/src/snitch_cluster_peripheral/snitch_cluster_peripheral.sv#L92-L95

zarubaf:
Yes exactly! We didn't have them before but now that have this event selector notion I think we can add them as they shouldn't come with significant overhead.

PR #255, opened 2021-09-15
periph: Performance counters

fischeti:
This PR integrates the existing performance counters into the runtime. The DMA performance metrics can now also be tracked (#247). Further, it fixes an issue where the `hwext` registers were not generated correctly by the `regtool`. It also adds the capability to reset counters at runtime.

### ToDo:
- [x] integrate `icache` stall and refill counters

### DIFF
--- a/hw/ip/snitch_dma/src/axi_dma_perf_counters.sv
+++ b/hw/ip/snitch_dma/src/axi_dma_perf_counters.sv
@@ -12,7 +12,8 @@ module axi_dma_perf_counters #(
     parameter int unsigned DATA_WIDTH         = -1,
     parameter type         axi_req_t          = logic,
     parameter type         axi_res_t          = logic,
-    parameter type         dma_events_t       = logic
+    parameter type         dma_events_t       = logic,
+    localparam bit         EnablePerfCounters = 0
 ) (
     input  logic                         clk_i,
     input  logic                         rst_ni,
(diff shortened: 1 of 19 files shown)


hw/ip/snitch_dma/src/axi_dma_perf_counters.sv
```diff
@@ -46,110 +51,142 @@ module axi_dma_perf_counters #(
 
         // defualt: keep old value
         dma_perf_d = dma_perf_q;
+        dma_events = '0;
 
         // aw
         if ( axi_dma_req_i.aw_valid) begin
           dma_perf_d.aw_valid_cnt = dma_perf_q.aw_valid_cnt + 'h1;
```
zarubaf, line 58: I don't think we want this, right? That gives a register for each perf counter (so will get quite big).

zarubaf (changes requested):
Looks good to me. Only thing is that we probably don't want to have the other counters in there as well, right? Maybe making it parameterizable or so.

hw/ip/snitch_dma/src/axi_dma_perf_counters.sv
fischeti: You mean the `dma_perf_q`? Those are the old performance counters and are not connected to anywhere, so I assume they will not be synthesized. But I can also delete or parametrize them

zarubaf: True, yes maybe a parameter would make the indent a but clearer. Then we are good to go!

zarubaf (approved):
Awesome @fischeti, super nice infrastructure! LGTM you can merge whenever you are ready :-)

PR #258, opened 2021-09-16
cva6: Add temporal fence

niwis:
Experimental feature for some security research.

- [x] Add experimental `fence.t` instruction and two CSRs (`fence_t_pad`, `fence_t_ceil`)
- [x] Add `apb_timer` peripheral to enable user-level timer required by seL4 (two timers, mapped to `0x2006000` and interrupt IDs `67` - `70`)
- [x] seL4 boots and channel-bench runs
- [x] Linux still boots
- [ ] `fence.t` executes correctly

Currently, executing `fence.t` during an seL4 context switch may deadlock. Further investigation is required.

_Note: these changes were verified on a version based on an older `master`: https://github.com/pulp-platform/snitch/tree/feature/fence-t. The observations above may not hold for this version._

Do you think we can merge nevertheless before the feature freeze? I do not expect big hardware changes to resolve the remaining issue with `fence.t` (if they still exist on the rebased version), and at least the tested version does not break anything.


zarubaf (approved):
Thanks @niwis!

PR #260, opened 2021-09-17
cluster: Add local CLINT-like software interrupt

noah95:
Adds a cluster-local CLINT-like software interrupt to replace the wakeup register. Atomic accecsses are guaranteed by a set/clear register pair. This solution supports only clusters of up to 32 cores.

## Banshee
The banshee wfi/wakeup was reworked. All relevant data structures are packed in the `WakeupState` struct which is protected by a mutex. The problem of harts exiting prematurely is alleviated by checking for any pending wakeups.

**There is a bug** however with the second test in `interrupt-local` (commented out): It can happen that a hart exits `wfi` but has its `mip` CSR not updated with the new `cl_clint` value and therefore doesn't jump to `mtvec`. This race condition occurs every ~10 executions on a fully loaded CPU with this command:

```bash
SNITCH_LOG=info banshee --num-cores=8 --num-clusters=4 --base-hartid=100 test-snRuntime-interrupt-local --trace > trace
```

All clint and wakeup accesses are atomic with `SeqCst` sequential consistency and I am out of ideas where this problem could come from. The `cl_clint` is loaded before each instruction by hart X, so after the `wfi` it reads `cl_clint` which has previously been `fetch_or`-ed by another hart Y and should have bit X set. This however sometimes is not the case 

```
; snrt_wfi (snrt.h:239)
;  in main (interrupt-local.c:81)
;  static inline void snrt_wfi() { asm volatile("wfi"); }
      800001dc                                          # wfi
; main (interrupt-local.c:82)
;  read_csr(mie);
      800001e0  x10=00080000                            # csrr    a0, mie
; main (interrupt-local.c:83)
;  read_csr(mip);
      800001e4  x10=00000000                            # csrr    a0, mip
```

Relevant functions:
- tran.rs: `emit_irq_check`
- engine.rs: `binary_wfi`, `wake`, `binary_store`

Perhaps @SamuelRiedel can spot the problem? Else, I would suggest merging anyway.


util/Makefrag
```diff
@@ -83,6 +83,12 @@ define QUESTASIM
 				-t 1ps -voptargs=+acc $1 +permissive-off ++$$1 \
 				-do "log -r /*; run -a"' >> $@
 	@chmod +x $@
+	@echo "#!/bin/bash" > $@.gui
+	@echo '${VSIM} +permissive -work ${VSIM_BUILDDIR} \
+				-ldflags "-Wl,-rpath,${FESVR}/lib -L${FESVR}/lib -lfesvr -lutil" \
+				-t 1ps -voptargs=+acc $1 +permissive-off ++$$1 \
+				-do "log -r /*; run -a"' >> $@.gui
+	@chmod +x $@.gui
```
zarubaf, line 91: Cool. Can you add this to the README please? Nice feature that people should be aware of :)

zarubaf (approved):
Great @noah95 - looks very good.

util/Makefrag
noah95: Good point, done in deb3f6a

SamuelRiedel (approved):

PR #265, opened 2021-09-20
banshee: Build banshee in container if changed

noah95:
Addressing #262

The `Get Banshee changed` job sets `banshee_changed` to `1` if there are any changes in `sw/banshee`. This re-builds banshee on pull requests that make changes to banshee. 

For workflows triggered on push, `github.event.pull_request.base.sha` is empty and banshee is not re-built


zarubaf (approved):
:+1: Terrific! Hopefully fixes those annoying failing CIs

SamuelRiedel (approved):
Awesome. Thanks for looking into this.

PR #267, opened 2021-09-23
runtime: Enable experimental extensions in clang

noah95:
Required by future LLVM toolchain that implements the smallfloat instructions


zarubaf (approved):
👍 Interesting that the one CI check fails (randomly).

noah95:
> one CI check fails (randomly).

please don't let it be some nasty race condition that occurs only once every decade. Although the RTL sim should produce the same every time..

PR #269, opened 2021-09-24
omp: Move to cluster-local clint

noah95:
Use the cluster-local clint for synchronization in OpenMP and further improvements

- L3 allocator can be placed at an offset to `_edram` since in the `cluster` runtime the putc-buffer is placed there, the L3 allocator has to back off a bit
- Rename sections in the linker script
- Store kmpc arguments in TCDM instead of global memory for performance


sw/snRuntime/include/omp.h
```diff
@@ -84,6 +85,13 @@ typedef struct {
      *
      */
     struct snrt_barrier *kmpc_barrier;
+    /**
+     * @brief Usually the arguments passed to __kmpc_fork_call would to a malloc
```
zarubaf, line 89:
```suggestion
     * @brief Usually the arguments passed to __kmpc_fork_call would do a malloc
```

sw/snRuntime/src/omp/kmp.c
```diff
@@ -14,16 +14,23 @@
 typedef void (*__task_type32)(_kmp_ptr32, _kmp_ptr32, _kmp_ptr32);
 typedef void (*__task_type64)(_kmp_ptr64, _kmp_ptr64, _kmp_ptr64);
 
-static void *args[sizeof(_kmp_ptr32)];
+/**
+ * @brief Usually the arguments passed to __kmpc_fork_call would to a malloc
```
zarubaf, line 18:
```suggestion
 * @brief Usually the arguments passed to __kmpc_fork_call would do a malloc
```

sw/snRuntime/src/omp/kmp.c
```diff
@@ -14,16 +14,23 @@
...
     kmp_int32 id = omp_get_thread_num();
-    _kmp_ptr32 *args = (_kmp_ptr32 *)arg;
-    kmpc_micro fn = (kmpc_micro)args[0];
     kmp_int32 *id_addr = (kmp_int32 *)(&id);
 
-    // first element in args in function pointer
-    _kmp_ptr32 *p_argv = &(args)[1];
+    // first element in args is the function pointer
+    kmpc_micro fn = (kmpc_micro)((_kmp_ptr32 *)arg)[0];
+    // second element in args in function pointer
```
zarubaf, line 32:
```suggestion
    // second element in args is the pointer to the argument vector
```

zarubaf (approved):
Awesome (as always 😄)

PR #272, opened 2021-09-27
RTL sim: fix return value overwrite with 0

noah95:
The return value stored in `a0` in RTL sims was overwritten in `_snrt_cluster_barrier`.

This PR further smuggles in the `make annotate` target in cluster sims. It generates source-code annotated traces from RTL simulation using the `annotate.py` script.


zarubaf (approved):
Awesome!

GiannaP (approved):
Looking good. Thanks Noah for taking care of this! :)

PR #268, opened 2021-09-23
Bugfix: fp encodings and banshee fp vector comparison instruction

GiannaP:
# Fixes for FP instructions
- instruction encoding for vfdotpex, vsumex, and vsum
- Banshee: bugfix of vector comparsions and adapting it to the latest HW Fpnew changes of integer registerfile writebacks

PR #273, opened 2021-09-27
fpnew: Fix Vivado flow

niwis:
Fix the Vivado build flow. In particular:
* apply the cva6 `fpu_wrap` patch added but reverted in #268,
* fix a typo in the patch filename from #268 (s.t. it is recognised by vendor),
* work around the following Vivado bug by realising the bus assignment in a bit-wise for-loop:
```
[Synth 8-523] part-select [15:0] does not match declaration [fpnew_sdotp_multi.sv:211]
```


GiannaP:
FYI: I just pushed a new update for the dotp unit: https://github.com/pulp-platform/snitch/tree/feature/upgrade-dotp
hopefully this does not cause duplicate work... :)

fischeti (approved):
Thanks for fixing this!

GiannaP (approved):
Looking good from my side :) As discussed in person, I have pushed the vivado fix to FPnew in  #273 so whoever merges last should clean up

PR #275, opened 2021-09-27
FPnew: upgrade dotp unit and bugfix

GiannaP:
New features / bugfixes:
- Upgrade the dotp unit of the FPnew to a version which has higher internal precision (almost fully exhaustivly tested with smaller examplary formats)
- upstreamed vivado fix to FPnew:  #273
- fix input operand width for vsum.s instruction


hw/vendor/pulp_platform_fpnew/src/fpnew_pkg.sv
```diff
@@ -61,7 +61,7 @@ package fpnew_pkg;
   typedef logic [0:NUM_FP_FORMATS-1][31:0] fmt_unsigned_t; // Unsigned indexed by FP format
 
   localparam fmt_logic_t CPK_FORMATS  = 6'b110000; // FP32 and FP64 can provide CPK only
-  localparam fmt_logic_t DOTP_FORMATS = 6'b001111; // FP32 and FP64 cannot be provided for DOTP
+  localparam fmt_logic_t DOTP_FORMATS = 6'b101111; // FP32 and FP64 cannot be provided for DOTP
```
noah95, line 64: Outdated comment?

GiannaP: ah good catch, it is a small hack to get wide enough operands for non-expanding `vsum.s`

PR #276, opened 2021-09-28
banshee: Fix `fcvt` + add `fmv` for half-floats

fischeti:
* Fixes `fcvt.s.h` instruction in banshee which was broken due to a `float` and `i32` mismatch.
* Adds `fmv.x.h` and `fmv.h.x` instructions

### DIFF
--- a/sw/banshee/flexfloat/src/lib.rs
+++ b/sw/banshee/flexfloat/src/lib.rs
@@ -325,7 +325,7 @@ pub unsafe fn ff_instruction_cvt_to_s(
     op: FfOpCvt,
     fpmode_src: bool,
     _fpmode_dst: bool,
-) -> f32 {
+) -> i32 {
     let env_fp8_src: flexfloat_desc_t = if fpmode_src { env_fp8alt } else { env_fp8 };
     let env_fp16_src: flexfloat_desc_t = if fpmode_src { env_fp16alt } else { env_fp16 };
 
@@ -351,8 +351,8 @@ pub unsafe fn ff_instruction_cvt_to_s(
         _ => (),
     };
 
-    let rd = ff_get_float(ff_res);
-    rd as f32
+    let rd = flexfloat_get_bits(ff_res);
+    rd as i32
 }
 
 /// conversions to fp64
--- a/sw/banshee/flexfloat/src/lib.rs
+++ b/sw/banshee/flexfloat/src/lib.rs
@@ -325,7 +325,7 @@ pub unsafe fn ff_instruction_cvt_to_s(
     op: FfOpCvt,
     fpmode_src: bool,
     _fpmode_dst: bool,
-) -> u32 {
+) -> f32 {
     let env_fp8_src: flexfloat_desc_t = if fpmode_src { env_fp8alt } else { env_fp8 };
     let env_fp16_src: flexfloat_desc_t = if fpmode_src { env_fp16alt } else { env_fp16 };
 
@@ -351,8 +351,8 @@ pub unsafe fn ff_instruction_cvt_to_s(
         _ => (),
     };
 
-    let rd = flexfloat_get_bits(ff_res);
-    rd as u32
+    let rd = ff_get_float(ff_res);
+    rd as f32
 }
 
 /// conversions to fp64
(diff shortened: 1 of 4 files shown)


SamuelRiedel (commented):
Looks good. I am not sure how the API between LLVM and Rust handles float types and whether the LLVM API will interpret the bits as a float or convert the value to a float.

In any case, if you have to change the Rust types to float here, don't you also have to change them for all the other functions converting flexfloat numbers? Like the `ff_instruction_cvt_to_d` function?

fischeti:
> In any case, if you have to change the Rust types to float here, don't you also have to change them for all the other functions converting flexfloat numbers? Like the `ff_instruction_cvt_to_d` function?

You are actually right, the instructions that rely on `ff_instruction_cvt_to_d ` also seem to be broken. I will fix them as well

fischeti:
Okay, the problem was different from what i thought. I changed the types back to `i32` to be more consistent (and also changing back `ff_get_float` to `flexfloat_get_bits` which returns an `i32`).

Instead, the problem was that `write_freg32` expects a `float` but received a `i32` which I fixed now with casting

SamuelRiedel:
> Okay, the problem was different from what i thought. I changed the types back to `i32` to be more consistent (and also changing back `ff_get_float` to `flexfloat_get_bits` which returns an `i32`).

Ok makes sense. However, we should still change all functions to use signed integers instead of unsigned then right? Because the [runtime](https://github.com/pulp-platform/snitch/blob/955ecf50b14ef80c764bb7466e4525f2fd91c1ee/sw/banshee/src/runtime/jit.ll#L47) uses signed everywhere. Or we could change the runtime.

GiannaP:
Two comments from my side:

**1)** Regarding the NaN-boxing (padding with 1): I oriented on the [SmallFloat documention](https://iis-git.ee.ethz.ch/smach/smallFloat-spec/-/blob/master/smallFloat_isa.pdf) of Stefan. In e.g. Chapter 1.2 it says:

> The NaN-boxing scheme described in Section 9.2 of [1] need not be enforced for half-precision
input operands in case the Xf16alt non-standard extension (see Chapter 2) is supported and must
not be performed for any floating-point input operands if the Xfvec non-standard extension (see
Chapter 4) is supported.

> NaN-boxing loses its usefulness when the alternative half-precision floating-point format is sup-
ported alongside with either half-precision (from Xf16) or quarter-precision (from Xf8) floating-
point formats. The goal of preventing register contents in narrower floating-point formats to
be read as valid operands for wider floating-point instructions breaks down in these cases On
one hand, since binary16 and binary16alt values have the same width, they can represent
valid operands in the respective other format. On the other hand, NaN-boxed binary8 values
represent valid binary16alt values since the exponent field in binary16alt overlaps the center
point of its 16-bit container.
Checking input operands for proper NaN-boxing when also supporting vectored floating-point
formats causes loss of efficiency when trying to perform scalar operations on the first entry of
vectors. Such cases would require masking other entries such that the first entry becomes properly
NaN-boxed again. Hence we decided against enforcing NaN-boxing when the Xfvec extension is
active.

The same is stated for other "small" formats.

@fischeti Does the HW NaN-box these instructions? I would say lets implement it the same way as the RTL works for keeping consistency between RTL and Banshee.

**2)** There are functions like `write_freg_f16` which are used for this (less lines of replicated code) and into which the NaN boxing should be moved in case HW does apply NaN boxing and we decide to go for that aswell. Currently, there is a comment in these functions that NaN boxing is not enable due to Vector support:
`// with vectorization enabled -> non NaN boxing`

fischeti:
> Ok makes sense. However, we should still change all functions to use signed integers instead of unsigned then right? Because the [runtime](https://github.com/pulp-platform/snitch/blob/955ecf50b14ef80c764bb7466e4525f2fd91c1ee/sw/banshee/src/runtime/jit.ll#L47) uses signed everywhere. Or we could change the runtime.

For some reason, the runtime only allows signed integers and according to @GiannaP, some `flexfloat` functions experienced  issues with signed integers. We could theoretically try to change everything to unsigned, but this would require to test every instruction again for which we currently do not have any tests yet. Those tests are in the pipeline so I propose to change this in the next PR. Is that okay with you?

fischeti:
> @fischeti Does the HW NaN-box these instructions? I would say lets implement it the same way as the RTL works for keeping consistency between RTL and Banshee.

The hardware does NaN-box the conversion instructions so I implemented it the same way in Banshee. I added the NaN boxing to the `write_freg_16` and `write_freg_f8`.

SamuelRiedel (approved):

GiannaP (approved):

PR #277, opened 2021-09-28
occamy: Generate c defines for soc peripheral registers

noah95:


GiannaP:
Does anyone have another input here (check my comment) or can we merge this?

PR #278, opened 2021-09-28
cluster: Verilator testbench with LLVM12

noah95:
The main issue for it not working out-of-the-box was the use of `ucontext` in the front-end server (fesvr). Accoring [to the Linux man page](https://linux.die.net/man/3/swapcontext), this cooperative scheduling technique is removed in *POSIX.1-2008* and POSIX threads are recommended. The patch takes care of that. 

Unfortunately, i couldn't notice significant speedup in either verilate- nor run-time. I will try to make the snitch cluster hierarchically compileable and multithreaded to see if that speeds up the compilation.

## Further changes
- Disable VCD output per default, can be enabled with `--vcd` when runnint `./bin/snitch_cluster.vlt`
- Distinct build location for `fesvr` from vsim and verilator since it might linked against different libraries which would break vsim/vlt simultaneous simulation
- For LLVM: use lld linker

### DIFF
--- a/util/Makefrag
+++ b/util/Makefrag
@@ -3,7 +3,6 @@ BENDER		   ?= bender
 DASM 	       ?= spike-dasm
 VLT			   ?= verilator
 
-# VERILATOR_ROOT ?= $(dirname $(shell which $(VLT)))/../share/verilator
 VERILATOR_ROOT ?= $(dir $(shell which $(VLT)))/../share/verilator
 VLT_ROOT	   ?= ${VERILATOR_ROOT}
 
@@ -22,6 +21,7 @@ FESVR          ?= ${MKFILE_DIR}work
 FESVR_VERSION  ?= 35d50bc40e59ea1d5566fbd3d9226023821b1bb6
 
 VLT_BUILDDIR := work-vlt
+VLT_FESVR     = $(VLT_BUILDDIR)/riscv-isa-sim
 VLT_FLAGS    += -Wno-BLKANDNBLK
 VLT_FLAGS    += -Wno-LITENDIAN
 VLT_FLAGS    += -Wno-CASEINCOMPLETE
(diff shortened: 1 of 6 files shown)


util/Makefrag
```diff
@@ -56,6 +86,22 @@ work/lib/libfesvr.a: work/${FESVR_VERSION}_unzip
 	mkdir -p $(dir $@)
 	cp $(dir $<)libfesvr.a $@
 
+# Build fesvr seperately for verilator since this might use different compilers
+# and libraries than modelsim/vcs and
+$(VLT_FESVR)/${FESVR_VERSION}_unzip:
+	mkdir -p $(dir $@)
+	wget -O $(dir $@)/${FESVR_VERSION} https://github.com/riscv/riscv-isa-sim/tarball/${FESVR_VERSION}
+	tar xfm $(dir $@)${FESVR_VERSION} --strip-components=1 -C $(dir $@)
+	patch -d $(dir $@) -p1 < ${ROOT}/util/patches/riscv-isa-sim/fesrv.patch
```
SamuelRiedel, line 95: Does this also work with gcc or is this something that should only happen with LLVM?

SamuelRiedel (approved):
Very cool that you got this to work!

zarubaf (approved):
LGTM, very nice re-factoring and features! 🎉

util/Makefrag
SamuelRiedel: Ah sorry, I was thinking about the patch you apply. But if the GCC flow still works with the patch then I am more than happy :)

noah95: This should clarify:

|      | gcc | llvm |
|------|-----|------|
| vsim | `ucontext` | not implemented |
| vlt  | `pthread` | `pthread` |

SamuelRiedel: Ah perfect. Thanks :)

PR #279, opened 2021-09-29
vendor: Remove deprecated opcodes file

noah95:
Can we remove `opcodes-sflt`? The `opcodes-flt-occamy` is the file used for smallfloat


noah95:
Since ariane conforms to the opcodes in `opcodes-sflt` we do not remove them but merge this PR with a notice.

PR #281, opened 2021-09-29
ariane_pkg: Add new FP formats

niwis:
This adds missing floating-point formats to `ariane_pkg`

- [x] Update `ariane_pkg`
- [ ] Check Linux boot on FPGA
- [ ] Update decoder


hw/vendor/openhwgroup_cva6/src/fpu_wrap.sv
```diff
@@ -65,7 +65,7 @@ module fpu_wrap import ariane_pkg::*; (
                  '{default: LAT_DIVSQRT},         // DIVSQRT
                  '{default: LAT_NONCOMP},         // NONCOMP
                  '{default: LAT_CONV},            // CONV
-                 '{default: fpnew_pkg::MERGED}},  // DOTP
+                 '{default: LAT_DOTP}},           // DOTP
       UnitTypes: '{'{default: fpnew_pkg::PARALLEL}, // ADDMUL
                    '{default: fpnew_pkg::MERGED},   // DIVSQRT
                    '{default: fpnew_pkg::PARALLEL}, // NONCOMP
```
GiannaP, line 71: Quesiton: Do we need the dotp unit in ariane? or can we put it to `fpnew_pkg::DISABLED`?

niwis: I think we do not need it (we did not need it so far), but is there a good reason not to include it? Is the overhead significant? Otherwise, this might be a "free" bonus feature

niwis: As discussed in person, it does not make much sense to support `DOTP` w/o the other FP formats and decoder support. I have disabled it in https://github.com/pulp-platform/snitch/pull/281/commits/22d718fe9bcf91d72121788f21f273c2e4d8f79e.

niwis:
OpenSBI successfully executes without any deadlocks 🎉 Linux still does not fully boot, but it gets much further than before this PR. (It seems to execute much slower than ~3 weeks ago, though)

GiannaP (approved):
Looking good :)

hw/vendor/openhwgroup_cva6/src/fpu_wrap.sv
```diff
@@ -54,7 +54,7 @@ module fpu_wrap import ariane_pkg::*; (
       Width:         64,
       EnableVectors: ariane_pkg::XFVEC,
       EnableNanBox:  1'b1,
-      FpFmtMask:     {RVF, RVD, XF16, XF8, XF16ALT},
+      FpFmtMask:     {RVF, RVD, XF16, XF8, XF16ALT, XF8ALT},
       IntFmtMask:    {XFVEC && XF8, XFVEC && (XF16 || XF16ALT), 1'b1, 1'b1}
```
GiannaP, line 58: Ah wait, a detail is missing:
```
IntFmtMask:    {XFVEC && (XF8 || XF8ALT), XFVEC && (XF16 || XF16ALT), 1'b1, 1'b1}
```

GiannaP (changes requested):
Sorry, overlooked that...

hw/vendor/openhwgroup_cva6/src/fpu_wrap.sv
niwis: Done in https://github.com/pulp-platform/snitch/pull/281/commits/96223ed1b56c1026e74eaa23aedcd16ecb0e85dc, and rebased onto latest `master`

GiannaP (approved):

GiannaP:
> OpenSBI successfully executes without any deadlocks tada Linux still does not fully boot, but it gets much further than before this PR. (It seems to execute much slower than ~3 weeks ago, though)

any idea what might be the issue for Linux not yet booting, or why it might be slower now?

PR #282, opened 2021-09-29
occamy: De-isolate snitches during startup

fischeti:
Snitch cores are isolated at start-up by default. The manager core de-isolates the Snitch cores in the beginning before trapping himself again. That is the current execution model to run 32bit applications on the Snitch clusters.

Additional changes:
* Fix `num_clusters` which was wrongly initialized
* Update `README` with some additional information to get started


zarubaf (approved):
👍

niwis (approved):
Looks good to me, I like the clean separation between the snitch and CVA6 bootroms. Perhaps we should at some point align this with the Linux flow

PR #286, opened 2021-10-04
Docker: Update from `snitch-llvm` to `pulp-llvm`

zarubaf:
Untested, but I guess we will see as soon as we merge.


noah95 (approved):
Thanks for that, didnt notice. Shoud be correct (see https://github.com/pulp-platform/llvm-project/releases/tag/0.1.0)

PR #288, opened 2021-10-05
Double the read-only cache's cache line width

SamuelRiedel:
Increase the read-only cache's line width and the number of supported address regions.

- [x] The latest version of Verible seems to screw up the formatting of structs that have comments in them and introduces a lot of whitespace changes. Which version of Verible are you using? Does someone have the CI's version 0.0-807-g10e7c71 working at ETH? Because this version does not work on my machine...

### DIFF
--- a/hw/system/occamy/src/memories.json
+++ b/hw/system/occamy/src/memories.json
@@ -28,6 +28,20 @@
         "words": 128
     },
     {
+        "byte_enable": false,
+        "byte_width": 8,
+        "density_optimized": true,
+        "description": [
+            "ro cache tag"
+        ],
+        "dual_port": false,
+        "latency": 1,
+        "ports": 1,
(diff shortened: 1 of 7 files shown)


zarubaf:
That should be better documented:

https://github.com/pulp-platform/snitch/blob/9be73c476e6667299eb3df14f6bfd25fff5d1fe4/.github/workflows/lint.yml#L10

zarubaf (approved):
LGTM. Mostly generated code anyway, right? Did the generation work?

SamuelRiedel:
Thanks, I saw that the CI uses this version, but it didn't work on my machine initially. I got it to work now and all the whitespace changes are gone.

And yes, the only 'real' change is in the `occamy_cfg.hjson`, the rest is generated and that worked fine.

PR #289, opened 2021-10-05
FPGA: Fix timing

niwis:
This fixes several timing violations for the FPGA flow:
- Reduce system clock from 50MHz to 25MHz (to fix a setup violation in the FPU)
- Reduce RTC from 25MHz to 7.5MHz
- Set RTC edge synchroniser as false path
- Remove false connection of Xilinx HBM port

With these changes, occamy compiles on FPGA without any timing violations.


zarubaf (approved):
Thanks @niwis. So the frequency drops for sure hurts a bit regarding boot time - but let's concentrate on functionality first :-)

PR #290, opened 2021-10-05
axi: Update data width converter to latest version

### LINKED ISSUE #283: axi: Outdated dw_converter
There is a bug in the current version of the data width converter that leads to requests being lost. This prevented me from accessing the soc control registers with Snitch. This bug is fixed in the [upstream AXI repo](https://github.com/pulp-platform/axi/commit/ada7c9e20c415c979d674f48d0e19ccdd0cf3f81#diff-5b1f86474f46437643d8849846d1f2c8e854c5058278f69c28cf5334a35f3427), but Snitch currently follows the DMA branch.

The solution would be to pull in the fixed dw_converter, but I am unsure how you want to handle this. Should we create a patch that only updates the dw_converter, or do you want to update the complete AXI repo?

SamuelRiedel:
A bug in the data width converter led to transactions to the soc_register being lost. This PR pulls in the latest version of the converter from the upstream AXI repo, which has the bug already fixed, as a patch.

Fixes #283


zarubaf (approved):
Thanks a lot @SamuelRiedel!

PR #298, opened 2021-10-14
occamy: Connect `chip_id` to SoC regs

fischeti:
This PR adds features which are used in the chip and system level testbenches of occamy

1. It connects the `chip_id` input in the top-level to the SoC regs such that it can be used by SW
2. Adds the possibility to disable the binary preloading in `fesvr`. This can be done by specifying the `--disable_preloading` flag. The use case for this is to test preloading over actual interfaces e.g. Serial link. `fesvr` preloading is still enabled by default


paulsc96 (approved):
LGTM, too :)

PR #308, opened 2021-11-04
occamy: Update cluster-cached regions, fix DRAM region

paulsc96:
This MR:

* Fixes a bug in cluster wrapper generation wherein the DRAM _length_ was used as a _mask_ of the corresponding cacheable region. This also fixes a related TODO by checking that the DRAM cached region is properly aligned for masking.
* Changes the cacheable regions in Snitch clusters in Occamy, which were stubbed with the above mistake. Now:
  * As in CVA6, the HBM and scratchpad are cached while the HBI is not.
  * Unlike CVA6, the boot ROM is also cached, massively speeding up boot and simulations by reducing access congestion. Since Snitch clusters are replicated and essentially worthless with a damaged instruction cache, this is an appropriate tradeoff, unlike in CVA6.
  
The final point is arguable. If there is disagreement, we can also keep the boot ROM uncached.


noah95:
Does the [`is_inside_cacheable_regions`](https://github.com/pulp-platform/snitch/blob/716efb16c55fe5eac3c2bfbf57f9177f9e8a9ec0/hw/ip/snitch/src/snitch_pma_pkg.sv#L58-L64) function work correctly for more than one cacheable region since [NrMaxRules is set to `1`](https://github.com/pulp-platform/snitch/blob/716efb16c55fe5eac3c2bfbf57f9177f9e8a9ec0/hw/ip/snitch/src/snitch_pma_pkg.sv#L10)? I probably oversee something but just to check

```sv
  localparam int unsigned NrMaxRules = 1;
  function automatic logic is_inside_cacheable_regions (snitch_pma_t cfg, logic[47:0] address);
    automatic logic [NrMaxRules-1:0] pass;
    pass = '0;
    for (int unsigned k = 0; k < cfg.NrCachedRegionRules; k++) begin
      pass[k] = range_check(cfg.CachedRegion[k].base, cfg.CachedRegion[k].mask, address);
    end
    return |pass;
  endfunction : is_inside_cacheable_regions
```

paulsc96:
> Does the [`is_inside_cacheable_regions`](https://github.com/pulp-platform/snitch/blob/716efb16c55fe5eac3c2bfbf57f9177f9e8a9ec0/hw/ip/snitch/src/snitch_pma_pkg.sv#L58-L64) function work correctly for more than one cacheable region since [NrMaxRules is set to `1`](https://github.com/pulp-platform/snitch/blob/716efb16c55fe5eac3c2bfbf57f9177f9e8a9ec0/hw/ip/snitch/src/snitch_pma_pkg.sv#L10)? I probably oversee something but just to check

Hehe, I believe not. Good catch! I remember having to adapt this before.

This means we probably have to do a bit more work here, since this parameter is hardcoded.

noah95:
> * Unlike CVA6, the boot ROM is also cached, massively speeding up boot and simulations by reducing access congestion. Since Snitch clusters are replicated and essentially worthless with a damaged instruction cache, this is an appropriate tradeoff, unlike in CVA6.

Since I'm also simulating Occamy chip I fully support caching the bootrom in Snitch :laughing:

paulsc96:
I gave up on trying to properly parameterize `NumMaxRules`, for now anyway. I just increased it statically to 4 in `snitch_pma_pkg`.

I have a macro-based solution that works with some tools, but not all apparently support hierarchical accesses to parameter types in member functions.

@colluca Tell me if you experience any issues with this solution, or whether it works for you.

hw/ip/snitch_cluster/src/snitch_cluster_wrapper.sv.tpl
```diff
@@ -100,9 +100,10 @@ package ${cfg['pkg_name']};
   localparam snitch_pma_pkg::snitch_pma_t SnitchPMACfg = '{
       NrCachedRegionRules: ${len(cfg['pmas']['cached'])},
       CachedRegion: '{
-% for cp in cfg['pmas']['cached']:
-          '{base: ${to_sv_hex(cp[0], cfg['addr_width'])}, mask: ${to_sv_hex(cp[1], cfg['addr_width'])}}${', ' if not loop.last else ''}
+% for i, cp in enumerate(cfg['pmas']['cached']):
```
noah95, line 103: Perhaps add a guard here to check for the hard-coded `4` in `snitch_pma_pkg.sv` while we're at it?

noah95 (changes requested):

noah95 (approved):

PR #311, opened 2021-11-10
occamy: Add appropriate AXI cuts in SoC

paulsc96:
This PR increases the number of additional spill register cuts on SoC-level AXI links to the following configuration:

* HBI <-> (Wide-X, Quadrants): 6
* Wide-X <-> (Quadrants, HBM): 3
* Narrow-X <-> (Quadrants, Peripherals): 3
* Narrow-X <-> (CVA6, Scratchpad): 1

### DIFF
--- a/hw/system/occamy/src/occamy_soc.sv
+++ b/hw/system/occamy/src/occamy_soc.sv
@@ -247,26 +247,6 @@ module occamy_soc
       .default_mst_port_i   ('0)
   );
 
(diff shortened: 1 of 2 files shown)


GiannaP (approved):
looking good! :)

paulsc96:
I will note here that we can probably ger away with smaller number of cuts for the final design. However, this is best addressed in a future PR.

PR #313, opened 2021-11-10
icache: Code improvements

SamuelRiedel:
Two minor code improvements in the icache:

- Change a duplicated, hardcoded value into a parameter
- Fix the refill in the lookup for a cache with a single set. This is a performance bug where we ignored half of the refills in such a configuration. AFAIK, we always have at least two sets at the moment, so nothing should change.


paulsc96 (approved):
LGTM

paulsc96 (changes requested):
Whoops. Let's add a `localparam` type actually, then we are good.

SamuelRiedel:
> Whoops. Let's add a `localparam` type actually, then we are good.

Good point. I fixed the linter warnings.

PR #315, opened 2021-11-19
axi: Fix `axi_id_remap` prioritizing ATOPs over AR channel

colluca:
Atomic swap operation (amoswap) requests on the AW channel (which have both write and read effects) are prioritized over requests on the AR channel. If a constant stream of amoswaps is present, this can stall the AR channel indefinitely possibly creating the conditions for a deadlock. 

This PR implements a very simple but fairer arbitration mechanism.


accuminium:
Rebased into single commit to clean up. Ready to merge from my side.

paulsc96 (approved):
LGTM

PR #316, opened 2021-11-26
snRuntime: Add test-and-test-and-set lock

colluca:
Add implementation of a test-and-test-and-set lock to snRuntime.

### DIFF
--- a/sw/snRuntime/include/snrt.h
+++ b/sw/snRuntime/include/snrt.h
@@ -249,13 +249,13 @@ static inline void snrt_wfi() { asm volatile("wfi"); }
  */
 static inline void snrt_mutex_lock(volatile uint32_t *pmtx) {
     asm volatile(
-        "li            x5,1          # x5 = 1\n"
+        "li            t0,1          # t0 = 1\n"
         "1:\n"
-        "  amoswap.w.aq  x5,x5,(%0)   # x5 = oldlock & lock = 1\n"
-        "  bnez          x5,1b      # Retry if previously set)\n"
+        "  amoswap.w.aq  t0,t0,(%0)   # t0 = oldlock & lock = 1\n"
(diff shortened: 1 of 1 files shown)


noah95:
Two small things regarding the code
- Could you change x5 to t0? Makes it more clear and ABI agnostic (really just a nit-pick)
- releasing the lock is done with the existing relase method?

General: This only works if there is no D$ right? And out of curiosity, what is the gain of flooding the interconnect with regulat loads than amos? Do you have performance numbers?

LucaColagrande-synthara:
> * Could you change x5 to t0? Makes it more clear and ABI agnostic (really just a nit-pick)

I used the same register as in the previous implementation. Shall I change both in that case?
Should we also change the name of the previous implementation to `snrt_mutex_tas_lock`?

> * releasing the lock is done with the existing relase method?

Yes.

> 
> General: This only works if there is no D$ right? 

Yes, or if the system is cache-coherent. Then again, what would work if the system was not cache-coherent?

> And out of curiosity, what is the gain of flooding the interconnect with regulat loads than amos?

As far as I understand, the AXI protocol can serve read and write requests concurrently. The `amoswap` is propagated on the write channel, but necessarily blocks the read channel as well, which is what was actually leading to the issue mentioned in PR https://github.com/pulp-platform/snitch/pull/315. Therefore, if the memory can serve reads and writes in parallel, I can see, in the best case, the TTAS lock reducing the bandwidth utilization by a factor of two.

> Do you have performance numbers?

One of the tests I have which does massive use of locks goes from 7527140 ns to 7201640 ns (in RTL simulation) when using TTAS instead of TAS. I know these numbers alone don't explain much, but there is a noticeable improvement.

fischeti (approved):
LGTM, I leave it up to you to add the suggestions from @noah95.

PR #318, opened 2021-11-27
occamygen.py: Add feature to generate generic C headers

colluca:
#### What
`occamygen.py` provides an option to generate a C header containing the system address map.
This PR adds an option to generate generic C headers from templates.

#### Why
This is important to write configuration independent software, which need not be manually edited when changing configuration.

#### Demo
As an example, the following C header can be generated:
```
#define NQUADS 8
#define NCLUSTERS 4
#define NCORES 9
```
from the following template [occamy_cfg.h.tpl]:
```
#define NQUADS ${cfg['nr_s1_quadrant']}
#define NCLUSTERS ${cfg['s1_quadrant']['nr_clusters']}
#define NCORES ${cfg['cluster']['nr_cores']}
```
by executing the following command:
```
occamygen.py -c occamy_cfg.hjson --cheader occamy_cfg.h.tpl
```

### DIFF
--- a/util/occamygen.py
+++ b/util/occamygen.py
@@ -77,11 +77,14 @@ def main():
     parser.add_argument("--bootdata",
                         metavar="BOOTDATA",
                         help="Name of the bootdata file (output)")
+    parser.add_argument("--cheader",
+                        metavar="CHEADER",
+                        help="Name of the cheader file (output)")
 
     parser.add_argument("--graph", "-g", metavar="DOT")
     parser.add_argument("--memories", "-m", action="store_true")
     parser.add_argument("--wrapper", "-w", action="store_true")
-    parser.add_argument("--cheader", "-D", metavar="CHEADER")
+    parser.add_argument("--am-cheader", "-D", metavar="ADDRMAP_CHEADER")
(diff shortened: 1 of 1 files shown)


paulsc96:
This seems like a good idea to me, but I'd like to understand more about the motivation.

This template mechanism only has access to the generator's state, right? So its main utility would be to generate different, user-defined headers with subsets of this information, like a `cluster.h`, a `cva6.h`, or a `periph.h`?

One problem I see is that this cannot generate all needed headers, e.g. those built from regtool register files. We would need some unified approach to also generating those.

colluca:
Yes it only has access to the generator's state indeed. As an example I'm using it to generate a `occamy_cfg.h` file (the contents of which are essentially what I posted above), such that the HW can be generated in a different configuration and the SW need not be manually edited, e.g. say you have to iterate over the number of Snitch cores.

noah95:
Wouldn't it be good to check in your template `occamy_cfg.h.tpl` file too?

colluca:
I have it in the snitch-pd repo, as all other SW-related things at the moment. Going to merge it there for the moment being, then after tomorrow's meeting I'm going to take some time to reorganize things as discussed.

PR #320, opened 2021-12-14
Automatic Latex addressmap generation (and config refactoring)

GiannaP:
This PR includes various changes around `occamygen.py`:

- `occamy_cfg.hjson`: includes now all addresses etc.
-  `solder.py`: can write a csv file for addressmap
-  `occamygen.py`: 
   - reads the base address + address region length directly from `occamy_cfg.hjson`
   - extended address map with quadrants, clusters (containing tcdm plus periph region)
   - writes a csv file for addressmap
- `addmapgen.py`: new script which takes the addrmap.csv and generates a memory map figure with the help of a latex template


noah95:
Generating the address map in the CI and adding it to a page in `systems/occamy` would be cool :sunglasses:. Could be done by extending `addmapgen.py` with a template/generator to `Markdown` and adding a new page [here](https://github.com/pulp-platform/snitch/blob/409525cb7d759eac5134ed372a27bb9b432f6647/mkdocs.yml#L37)

hw/system/occamy/src/occamy_cfg.hjson
```diff
@@ -168,10 +154,102 @@
...
+        address_0: 2147483648, // 0x8000_0000
+        address_1: 68719476736, // 0x10_0000_0000
+        channel_size: 1073741824, // 1 GiB 0x4000_0000
+        nr_channels_total: 8,
+        nr_channels_address_0: 2,
+    },
+    // dram corresponds to 'hbm address_0' and 'nr_channels_address_0'
+    dram: { 
+        address: 2147483648, // 0x8000_0000
+        length: 2147483648, // 2 GiB 0x8000_0000
```
GiannaP, line 253: I am not sure if we should merge these configs somehow as the `dram` corresponds to `address_0` of `hbm` with `nr_channels_address_0` channels that are mapped into that memory region.

hw/system/occamy/src/occamy_cluster_wrapper.sv
```diff
@@ -77,8 +77,8 @@ package occamy_cluster_pkg;
       NrCachedRegionRules: 4,
       CachedRegion: '{
           0: '{base: 48'h80000000, mask: 48'hffff80000000},
-          1: '{base: 48'h1000000000, mask: 48'hfff000000000},
-          2: '{base: 48'h70000000, mask: 48'hfffff0000000},
+          1: '{base: 48'h1000000000, mask: 48'hfffe00000000},
+          2: '{base: 48'h70000000, mask: 48'hfffffffe0000},
```
GiannaP, line 81: @paulsc96 : In the `occamy.py` file I extract the base address and address region length newly directly from the occamy cfg. When adding the new areas, did you overprovision a bit when adding the address region length?

util/clustergen/occamy.py
```diff
@@ -20,12 +20,24 @@ def __init__(self, cfg):
...
+        pma_cfg.add_region_length(PMA.CACHED,
+                                  cfg["hbm"]["address_1"],
+                                  cfg["hbm"]["nr_channels_total"] * cfg["hbm"]["channel_size"],
+                                  addr_width)
         # Make the SPM cacheable
-        pma_cfg.add_region_length(PMA.CACHED, 0x7000_0000, 0x1000_0000, addr_width)
+        pma_cfg.add_region_length(PMA.CACHED,
+                                  cfg["spm"]["address"],
+                                  cfg["spm"]["length"],
+                                  addr_width)
```
GiannaP, line 35: changes caused RTL change: see other [comment](https://github.com/pulp-platform/snitch/pull/320/files?authenticity_token=5LqIqefCik%2BESGIAcu366MJA0sTfpx1qluazva1%2FbsiU1KJhz43fa%2Fi1KlYIQ67gr9gTmaYrVg7%2F%2BXwDOs0r7g%3D%3D&file-filters%5B%5D=.py&file-filters%5B%5D=.sv&file-filters%5B%5D=.tex&file-filters%5B%5D=dotfile#r769805452)

util/clustergen/occamy.py
```diff
@@ -20,12 +20,24 @@ def __init__(self, cfg):
...
-        pma_cfg.add_region(PMA.CACHED, 0x0000_8000_0000, 0xffff_8000_0000)
-        pma_cfg.add_region(PMA.CACHED, 0x0010_0000_0000, 0xfff0_0000_0000)
+        pma_cfg.add_region_length(PMA.CACHED,
+                                  cfg["hbm"]["address_0"],
+                                  cfg["hbm"]["nr_channels_address_0"] * cfg["hbm"]["channel_size"],
+                                  addr_width)
+        pma_cfg.add_region_length(PMA.CACHED,
+                                  cfg["hbm"]["address_1"],
+                                  cfg["hbm"]["nr_channels_total"] * cfg["hbm"]["channel_size"],
+                                  addr_width)
```
GiannaP, line 30: changes caused RTL change: see other [comment](https://github.com/pulp-platform/snitch/pull/320/files?authenticity_token=5LqIqefCik%2BESGIAcu366MJA0sTfpx1qluazva1%2FbsiU1KJhz43fa%2Fi1KlYIQ67gr9gTmaYrVg7%2F%2BXwDOs0r7g%3D%3D&file-filters%5B%5D=.py&file-filters%5B%5D=.sv&file-filters%5B%5D=.tex&file-filters%5B%5D=dotfile#r769805452)

hw/system/occamy/src/occamy_cluster_wrapper.sv
paulsc96: Indeed, the masks of those regions were manually overprovisioned in `util/clustergen/occamy.py::23` onward. This was done so that less bits have to be matched, since this matching is directly on the (near-critical) Snitch decode path. Shouldn't make much of a difference, but just something to keep in mind during implementation.

hw/system/occamy/src/occamy_cfg.hjson
```diff
@@ -168,10 +154,102 @@
         num_itlb_entries: 1,
     }
     // peripherals
+    debug: {
```
paulsc96, line 157: If you want to add configuration items to the config (which is a good idea), please also remember to adapt the documentation schema (`docs/schema/occamy.schema.json`); otherwise, the documentation does not match the config, which is a bad time for people getting started.

You will then also have to rebuild the doc, which is a bit of a pain. You can adapt the schema and I will push a build commit.

util/addrmapgen.py
```diff
@@ -0,0 +1,391 @@
...
+    if (max_nr_merged_entries % 2) == 0:
+        idx0 = int(half-1)
+        idx1 = int(half)
+    else:
+        idx0 = int(half)
+        idx1 = int(half)+1
+    return (idx0, idx1)
+
+
+def main():
```
paulsc96, line 69: Bit of a nitpick, but I prefer to parse the args in the `if __name__ == '__main__'` block and pass them to the main function unpacked; this gives `main()` (or whatever invoked function) a clear interface with (if type annotations are used) well-understood args.

util/clustergen/occamy.py
paulsc96: As mentioned above, to be seen if this affects/degrades implementation quality (hopefully not, since I prefer it on prinicple).

GiannaP:
> Generating the address map in the CI and adding it to a page in `systems/occamy` would be cool sunglasses. Could be done by extending `addmapgen.py` with a template/generator to `Markdown` and adding a new page [here](https://github.com/pulp-platform/snitch/blob/409525cb7d759eac5134ed372a27bb9b432f6647/mkdocs.yml#L37)

Good idea, I added a simple markdown table generation and added the markdown file as proposed in the documentation ;)

paulsc96 (changes requested):
Looks good overall, thanks for parameterizing and updating the stale generation code.

If you can update the Occamy config doc schema, I can regenerate the doc and we can merge :+1:

hw/system/occamy/src/occamy_cluster_wrapper.sv
paulsc96: I'd say we should go ahead with it for now and revisit it in case it causes problems. I prefer the approach of generating the exact mask, so if it turns out not a problem, all the better.

hw/system/occamy/src/occamy_cfg.hjson
GiannaP: Ah completely forgot the schema, of course, I will add them! Ideally, we would have a CI check to ensure they are kept up-to-date.

util/addrmapgen.py
GiannaP: Thanks for the input! I have no strong feelings on this, so I can implement your proposed changes.

paulsc96 (approved):
LGTM :)

PR #324, opened 2021-12-17
occamy: Generic peripheral configuration

GiannaP:
Various changes like:
- add new `peripherals.schema.json` which is newly an object of the `occamy.schema.json` which contains a variable-length array of peripherals for AXI LIte interconnect and for Regbus interconnect instead of the previous hardcoded peripherals.
- adapt `occamy_cfg.hjson` to the new schema
- adapt `occamy_gen.py` to use and understand the new schema / new config structure
- update documentation
- `occamy_pk.sv.tpl`: replace hardcoded values for `cluster_base_addr` and `cluster_base_offset` from `occamy_conf.hjson` via the `occamygen.py` generation script


hw/system/occamy/test/occamy.dts
```diff
@@ -42,12 +42,6 @@
     #size-cells = <2>;
     compatible = "simple-bus";
     ranges;
-    debug@0 {
```
paulsc96, line 45: The debug module description seems to have disappeared entirely from the device tree; can you double-check this?

paulsc96 (changes requested):
Looks good, except for the missing debug module in the device tree; something seems to have gone awry in the generator.

hw/system/occamy/test/occamy.dts
GiannaP: oh good catch! I will double check

niwis: Yes, the `debug` entry is deprecated: https://github.com/openhwgroup/cva6/pull/806

paulsc96: In that case, we should also remove it from the generator code.

Also, looking at the UART fix in that PR, this also directly concerns us, right @niwis?

niwis: Yes, indeed. I would cherry-pick the PR once it is merged.

GiannaP: I fixed it to the previous state (with debug module). The other CVA6 and linux booting related changes (remove debug and fix the UART) should be done in another PR for keeping it easier trackable.

paulsc96 (approved):

PR #327, opened 2022-01-20
occamy: Quadrant manager, hierarchical wide interconnect, and HBI reduction

paulsc96:


hw/system/occamy/src/occamy_soc_ctrl/occamy_soc_reg.hjson.tpl
```diff
@@ -101,6 +96,23 @@
...
+    { name: "NUM_QUADRANTS"
+      desc: "Number of quadrants per chip."
+      swaccess: "ro"
+      hwaccess: "none"
+      hwqe:     "true",
+      hwext:    "true",
+      fields: [
+        {
+          bits: "31:0"
+          resval: ${nr_s1_clusters}
```
GiannaP, line 108: Based on the name and description I would this expect to be `nr_s1_quadrant` instead of  `nr_s1_clusters`?

paulsc96: Good point. This is a typo :| I presume during my test I made some superfluous accesses that went nowhere. I will fix any issues in bulk later.

util/occamygen.py
```diff
@@ -437,13 +469,92 @@ def main():
...
+    # Default port: soc wide xbar
+    quadrant_inter_xbar.add_output_entry("wide_xbar", am_soc_wide_xbar)
+    quadrant_inter_xbar.add_input("wide_xbar")
+    for i in range(nr_s1_quadrants):
+        # Default route passes HBI through quadrant 0
+        # --> mask this route, forcing it through default wide xbar
+        quadrant_inter_xbar.add_output_entry("quadrant_{}".format(i),
+                                             am_wide_xbar_quadrant_s1[i],
+                                             range_mask=(0, 0x1000000000))
+        quadrant_inter_xbar.add_input("quadrant_{}".format(i))
```
GiannaP, line 529: A short clarification question: so by default it would have routed through Quadrant 0 to the NarrowXbar to the Narrrow Die2Die link?

GiannaP (approved):
This was quite some code to go through, but LGTM :) only some small comments/questions. Thanks a lot for your work!

util/occamygen.py
paulsc96: Aah, good catch. This is a leftover from when every quadrant had its own HBI port and the routing algorithm of solder would insist on routing all HBI traffic through quadrant 0 in the inter xbar (fewer hops). Since this is no longer possible, this fix is redundant now.

hw/system/occamy/src/occamy_quadrant_s1.sv
```diff
@@ -361,14 +351,14 @@ module occamy_quadrant_s1
       .req_t(axi_a48_d512_i4_u0_req_t),
       .resp_t(axi_a48_d512_i4_u0_resp_t)
   ) i_wide_cluster_out_isolate (
-      .clk_i(clk_i),
-      .rst_ni(rst_ni),
-      .slv_req_i(snitch_ro_cache_req),
-      .slv_resp_o(snitch_ro_cache_rsp),
+      .clk_i(clk_quadrant),
+      .rst_ni(rst_quadrant_n),
```
GiannaP, line 355: wrong clk and reset :)

hw/system/occamy/src/occamy_quadrant_s1.sv
```diff
@@ -254,18 +247,18 @@ module occamy_quadrant_s1
       .req_t(axi_a48_d64_i4_u0_req_t),
       .resp_t(axi_a48_d64_i4_u0_resp_t)
   ) i_narrow_cluster_out_isolate (
-      .clk_i(clk_i),
-      .rst_ni(rst_ni),
+      .clk_i(clk_quadrant),
+      .rst_ni(rst_quadrant_n),
```
GiannaP, line 251: wrong clock, aswell :)

util/occamygen.py
```diff
@@ -455,33 +566,22 @@ def main():
...
-    soc_wide_xbar.add_input("soc_narrow")
+    # Default port: HBI (always escalate "upwards" in hierarchy -> off-chip)
+    soc_wide_xbar.add_output_entry("hbi", am_hbi)
+    soc_wide_xbar.add_output_entry("hbm_xbar", am_hbm_xbar)
+    soc_wide_xbar.add_output_entry("quadrant_inter_xbar", am_quadrant_inter_xbar)
     soc_wide_xbar.add_output_entry("soc_narrow", am_soc_narrow_xbar)
-
-    # TODO(zarubaf): PCIe should probably go into the small crossbar.
-    soc_wide_xbar.add_input("pcie")
-    soc_wide_xbar.add_output_entry("pcie", am_pcie)
```
noah95, line 476: In the FPGA designs we (mis-)use the PCI ports for peripherals (QSPI, HBM config, ...). Could we not remove this and make it parametrizable in `occamy_cfg.hjson`?

noah95 (changes requested):

util/occamygen.py
paulsc96: The PCI port is still present and now connected to the narrow X-bar. Do you need it on the wide x-bar? If so, you can instead use the single HBI port, which now goes directly to the wide X-bar; there is no difference between this new HBI connection and the old PCI connection.

hw/system/occamy/Makefile
```diff
@@ -81,16 +82,22 @@ update-plic:
 	@$(ROOT)/hw/vendor/lowrisc_opentitan/rv_plic/util/reg_rv_plic.py -s 70 -t 2 -p 6 $(ROOT)/hw/vendor/lowrisc_opentitan/rv_plic/data/rv_plic.hjson.tpl > src/rv_plic/rv_plic.hjson
 	@$(ROOT)/util/regtool.py -r -t src/rv_plic src/rv_plic/rv_plic.hjson
 
-# Regenerate the SoC register.
+# Update generic control register.
```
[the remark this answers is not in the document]
paulsc96: Done.

util/occamygen.py
noah95: didn't catch that, narrow is fine thanks!

PR #328, opened 2022-01-28
hw: Ensure consistent reset practices and strategy

paulsc96:
This ensures (hopefully) that all RTL in the repo, and particularly in the cluster, has a consistent reset style: asynchronous and (globally) active-low.

I would have preferred to verify the completeness of this in some analysis tool first, but since all of the changed registers need to be fixed anyways, we might as well merge this now and add any additional fixes later.


GiannaP (approved):
LGTM!

PR #329, opened 2022-01-31
ci: Fix unintended command execution in lint message

GiannaP:
The CI check `Occamy` in `.github/workflows/lint.yml` ran the following commands:

```bash
make -C hw/system/occamy all
util/git-diff.py --error-msg "::error ::Found differences, run `make -C hw/system/occamy` all before committing."
```
The \` in the argument to the `git-diff.py` script caused the make command enclosed by \` `make -C hw/system/occamy` to be executed which broke the `git-diff.py` script.


paulsc96 (approved):
Good catch! Thanks :+1:

PR #330, opened 2022-01-31
Fix syntax in occamy cluster and set NrMaxRules based on the number of defined cached regions

lucabertaccini:
The previous assignment pattern in `occamy_cluster_wrapper.sv` was not supported by some tools.
* This PR fixes that assignment
* `NrMaxRules` set in `snitch_pma_pkg` is now computed from the number of cached regions defined in `occamy.py` and `cluster.py`. Two separate files containing a `snitch_pma_pkg`, one for occamy (`occamy_snitch_pma_pkg.sv`) and one for snitch (`snitch_pma_pkg.sv`), are generated. --> lint checks complain about that
* A target was inserted in the `Bender.yml` file to select the right file containing a `snitch_pma_pkg` to be included in the scripts generated by Bender. The correspondent flag was added in the `Makefile` used to generate the occamy system

### DIFF
--- a/hw/ip/snitch/Bender.yml
+++ b/hw/ip/snitch/Bender.yml
@@ -19,7 +19,12 @@ export_include_dirs:
 
 sources:
 # Level 0:
-- src/snitch_pma_pkg.sv
+- target: not(occamy)
+  files:
+  - src/snitch_pma_pkg.sv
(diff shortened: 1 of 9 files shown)


paulsc96:
Obviated by #331.

PR #331, opened 2022-01-31
cluster: Fix memory region syntax in wrapper

lucabertaccini:
Fix an assignment pattern in the cluster wrapper that was not supported by some tools


paulsc96 (approved):
LGTM :+1:

PR #334, opened 2022-02-10
Add Zeromemory

GiannaP:
- add ZeroMemory module and attach it to the DMA AXI (wide 512bit) in the Snitch cluster
- adapt `cluster` schema/config to set the size of `cluster_peripheral` and `zero_mem` address region
- use memory region sizes when generate `snitch_cluster_wrapper` to set new parameters for configuring the address regions in RTL
- add `zero_mem` address region to memory map
- add SW test for the Zero Memory
- use occamygen to generate a C-header with all base addresses directly into the `snRuntime/include` directory


hw/ip/snitch_cluster/src/snitch_cluster.sv
```diff
@@ -415,7 +419,15 @@ module snitch_cluster
   // Calculate start and end address of TCDM based on the `cluster_base_addr_i`.
   addr_t tcdm_start_address, tcdm_end_address;
   assign tcdm_start_address = cluster_base_addr_i;
-  assign tcdm_end_address = cluster_base_addr_i + TCDMSize;
+  assign tcdm_end_address   = cluster_base_addr_i + TCDMSize;
```
paulsc96, line 422: As in #335, we should be sure that cluster-internal start and end addresses are masked in HW so that the LSBs of `cluster_base_addr_i` do not propagate. I believe the masking of `tcdm_start_address` and `tcdm_end_address` in #335 is also sufficient here, but better to think about it twice.

sw/snRuntime/src/dma.c
```diff
@@ -52,8 +52,8 @@ snrt_dma_txid_t snrt_dma_start_1d(void *dst, const void *src, size_t size) {
 
 /// Initiate an asynchronous 2D DMA transfer with wide 64-bit pointers.
 snrt_dma_txid_t snrt_dma_start_2d_wideptr(uint64_t dst, uint64_t src,
-                                          size_t size, size_t dst_stride,
-                                          size_t src_stride, size_t repeat) {
+                                          size_t size, size_t src_stride,
```
paulsc96, line 55: this change seems a bit counterintuitive; is there a reason why the strides should not have the same order as the addresses?

paulsc96 (approved):
LGTM; I have some comments, but I don't mind if we merge as-is either :)

hw/ip/future/src/axi_zero_mem.sv
```diff
@@ -0,0 +1,81 @@
...
+// Solderpad Hardware License, Version 0.51, see LICENSE for details.
+// SPDX-License-Identifier: SHL-0.51
+
+// Authors:
+// - Gianna Paulin <pauling@iis.ee.ethz.ch>
+
+`include "common_cells/registers.svh"
+/// AXI4+ATOP slave module which translates AXI bursts into a memory stream.
+/// If both read and write channels of the AXI4+ATOP are active, both will have an
+/// utilization of 50%.
```
fischeti, line 11: This is probably copy-pasted from the `axi_to_mem` module. Maybe add a short description about the zero-memory.

sw/snRuntime/src/dma.c
fischeti: There is an inconsistency between `snrt_dma_start_2d` and `snrt_dma_start_2d_wideptr`. The former calls the latter, but `dst_stride` and `src_stride` are swapped in the process. Maybe it makes more sense to change it in `snrt_dma_start_2d` and always take the `dst` arguments before `src` arguments to be consistent

sw/snRuntime/tests/zero_mem.c
```diff
@@ -0,0 +1,183 @@
...
+    uint32_t *buffer_tcdm = (void *)snrt_cluster_memory().start;
+    uint32_t *buffer_golden = (void *)(snrt_cluster_memory().start + 128);
+
+    // printf("Zero Memory at %p\n", zero_mem);
+    // printf("TCDM Memory at %p\n", buffer_tcdm);
+
+    ///////////////
+    // CORE READ //
+    ///////////////
+    if (snrt_is_compute_core()) {
```
fischeti, line 25: Nitpick, all compute cores in a cluster will execute this section. It should still function correctly. Could also be changed to `if (snrt_cluster_compute_core_idx() == 0)`

util/addrmapgen.py
```diff
@@ -213,7 +213,7 @@ def main(file: str, outdir: pathlib.Path, tex_filename: str, md_filename: str):
     nr_q_entries = len(all_quadrant_entries)
 
     # Quadrant latex
-    NR_CLUSTER_ADDR_RANGES = 2  # tcdm and periphs
+    NR_CLUSTER_ADDR_RANGES = 3  # tcdm and periphs
     NR_CLUSTERS_PER_QUADRANT = 4  # hardcoded for now
```
fischeti, line 217: Unrelated to this PR, but I noticed that the hardcoded `NR_CLUSTERS_PER_QUADRANT ` breaks the generator if you try out smaller configurations (I think because of zero divides). Would this be difficult to change?

sw/snRuntime/include/snrt.h
```diff
@@ -91,8 +91,12 @@ extern int snrt_is_compute_core();
 extern int snrt_is_dm_core();
 extern void snrt_wakeup(uint32_t mask);
 
+/// get start address of global memory
 extern snrt_slice_t snrt_global_memory();
+/// get start address of the cluster's tcdm memory
 extern snrt_slice_t snrt_cluster_memory();
+/// get start address of the cluster's zero memory
+extern snrt_slice_t snrt_zero_memory();
```
fischeti, line 99: This is nice!

fischeti (approved):
LGTM! I leave it up to you to add my suggestions or not.

fischeti:
One thing that just came to my mind, is that the zero-memory is not implemented in banshee. I think it would be quite handy to support e.g. `snrt_zero_memory()` also in banshee, because I normally start prototyping SW in banshee and then move to the RTL. Although I don't know the best way to implement this in banshee

hw/ip/snitch_cluster/src/snitch_cluster.sv
GiannaP: I think your approach from #335 should prevent it. Although I am wondering if I should maybe also mask the new addresses:
-  `cluster_periph_*_address`: goes into the barrier & AXI addrmap
- `cluster_zero_mem_*_address`: AXI addrmap

as they do not go into the `snitch_cc` etc. I think they are less critical and in theory I would expect them to be optimized, not sure though...

util/addrmapgen.py
GiannaP: Yeah, that was not tested. However, I think it should be a rather easy fix. Do you want to use it for the Dory trial setup? If so, do you have a branch where I could fix it?

PR #335, opened 2022-02-10
tcdm: Reduce address width, leverage base alignment

paulsc96:
This PR hopes to reduce the complexity and area of TCDM components by:
* Reducing the unnecessarily wide TCDM request addresses in clusters from `AddrWidth` to a mimimal `TcdmAddrWidth` (e.g. from 48 to 17 in Occamy).
* Ensuring in hardware that the `tcdm_start_address` and `tcdm_end_address` are hardware-masked (this is already enforced by assertions, but those do not influence logic synthesis).
* Removing the `tcdm_start_addr_i` input from the SSR streamers entirely, since their address space is now constrained exactly to the TCDM and the input is pointless.
* Simplifying SSR configuration to default on pointer widths covering TCDM exactly by default, and indices covering TCDM in a unit-stride 1D loop; this reduces index/pointer width defaults from 16/18 to 14/17 for 128 KiB TCDM.
* Adapting the SSR standalone testbenches to the new default widths.

The MR also fixes an assertion bug in the isochronous crossings in `hw/vendor/pulp_platform_common_cells`.


hw/ip/snitch_cluster/src/snitch_cluster.sv
```diff
@@ -577,6 +581,7 @@ module snitch_cluster
     .mem_rdata_i (ext_dma_rsp.p.data)
   );
 
+  assign ext_dma_req.q.addr = tcdm_dma_req_t'(ext_dma_req_q_addr_nontrunc);
```
GiannaP, line 584: Not sure if it works like that, but I would have gone with `tcdm_addr_t` instead of `tcdm_dma_req_t`?

GiannaP (commented):
LGTM :) just with the one type conversion I am a bit confused.

hw/ip/snitch_cluster/src/snitch_cluster.sv
paulsc96: Actually, that cast is mostly eyecandy here, indicating that we intend to truncate, but indeed, the type is wrong. It should be `tcdm_dma_addr_t`; will fix.

GiannaP (approved):

PR #337, opened 2022-02-11
sw: Applications & kernels

fischeti:
Adds applications and kernels to the Snitch `sw` directory. 

### Kernels
Kernels are hand-optimized for Snitch systems with SSR streams and `frep` loops. All kernels are written in double precision, some have additional versions for smaller floating point precision. The following kernels are available:
* GEMM, multiple floating points format (`double`,` float`, `__fp16`), sub 64-bit floating formats use SIMD instructions
* BatchNorm, implemented as a simple `y = x * kappa + gamma`
* MaxPool
* FusedConv, kernel that fuses 2D convolution + BatchNorm + ReLU. has an interface that is supported by DORY

### Layers
Each kernel is wrapped by a layer module, that handles DMA transfer in a double-buffered fashion

### SW testbenches
for each of the kernels, there is a testbench called `net_*` that allows profiling the performance of the kernel and verifying the functionality of the kernels.

### Data generation
Data header files for the kernels are generated by a python script `data_gen.py` which also generates checksums to for verification purposes. Dimensions for each kernel can be specified in config files. Data header files are automatically generated with CMake, but require the `torch`  python package.

### Misc
* Small fixes in `CMakeLists.txt` inside the `hw/system` folders that prevented overwritting the `SNITCH_RUNTIME` variable

## TODO:

- [x] pass arguments as structs by reference for all `conv2d` functions
- [x] Add to CI


GiannaP:
Quite some code to go through (I am not finished yet), but nice work! :)

I have already two questions/remarks:
- as you did some nice commenting, can we add it to [doxygen ](https://github.com/pulp-platform/snitch/blob/96b5ffb440a648aefa087978224dae3187b07e58/sw/Doxyfile#L15-L19) such that it will be added in the [documentation ](https://pulp-platform.github.io/snitch/runtime/Files/)
- is the usage (ordering) of the source and destination strides for `snrt_dma_start_2d` correct as is? (Since we changed it in #334)

sw/applications/include/layer.h
```diff
@@ -0,0 +1,73 @@
+// Copyright 2020 ETH Zurich and University of Bologna.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <stdint.h>
+
+typedef enum { CONV2D, GEMM, POOLING, BATCH_NORM } LAYER_TYPE;
```
noah95, line 9: Not sure if we adhere to any kind of C style guide, but i like to append any typedefs with `_t` and keep them lowercase. e.g. `layer_type_t` and `activation_t`.

sw/applications/include/layer.h
```diff
@@ -0,0 +1,73 @@
...
+
+typedef enum { RELU, SOFTMAX } ACTIVATION;
+
+typedef enum { FP64 = 8, FP32 = 4, FP16 = 2, FP8 = 1 } PRECISION;
+
+struct layer;
+typedef struct layer_struct layer;
+
+struct layer_struct {
```
noah95, line 18: This is a general layer struct right? So if we only want to run a `GEMM` layer we need this entire struct in memory with all the other members from conv2d and batchnorm that are not used? Could we split this into separate structs `struct layer_gemm` `struct layer_conv2d` etc?

sw/applications/src/kernels/batchnorm.c
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @param ifmap pointer to input feature map
+ * @param gamma pointer to gamma
+ * @param beta pointer to beta
+ * @param ofmap pointer to output feature map
+ * @param OW width of output feature map
+ * @param CI number of input channels
+ * @param compute_num number of compute units
+ * @param setup_SSR setup SSR strides and bounds
+ */
```
noah95, line 19: Duplicate docstring: If the header is documented we can omit this here and leave only in-line comments to document the code.

sw/applications/src/kernels/conv2d.h
```diff
@@ -0,0 +1,192 @@
...
+    const uint16_t ch_in, const double* pWeight, const uint16_t ch_out,
+    const uint16_t dim_kernel_x, const uint16_t dim_kernel_y,
+    const uint16_t padding_y_top, const uint16_t padding_y_bottom,
+    const uint16_t padding_x_left, const uint16_t padding_x_right,
+    const uint16_t stride_x, const uint16_t stride_y, const int8_t* bias,
+    const uint16_t bias_shift, const uint16_t out_shift,
+    const uint16_t out_mult, double* pOutBuffer, const uint16_t dim_out_x,
+    const uint16_t dim_out_y, double* k, double* lambda, double* pIm2ColBuffer,
+    int flag_relu, int flag_batch_norm, int flag_y_accumulate_start,
+    int flag_y_accumulate_end, unsigned int* memory_chan);
```
noah95, line 58: So many argument! They are all pushed/popped to/from stack increasing the risk of a stack-overflow and making the function call slower. This could be improved by declaring a struct and only pass the pointer to this struct.

Since it is `noinline` the compiler will not inline it and therefore also cannot optimize it.

sw/applications/src/kernels/conv2d.h
```diff
@@ -0,0 +1,192 @@
...
+
+void __attribute__((noinline)) occamy_conv_opt_fp32(
```
noah95, line 60: Missing docstring, could refer to `occamy_conv_opt_fp64`

sw/applications/src/kernels/gemm.c
```diff
@@ -0,0 +1,582 @@
+// Copyright 2020 ETH Zurich and University of Bologna.
```
noah95, line 1: again, duplicate doc-string. Just keep them in the header

sw/applications/src/kernels/maxpool.c
```diff
@@ -0,0 +1,30 @@
+// Copyright 2020 ETH Zurich and University of Bologna.
```
noah95, line 1: duplicate docstring

sw/applications/src/layers/batchnorm_layer.c
```diff
@@ -0,0 +1,136 @@
...
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "batchnorm_layer.h"
+
+#include "batchnorm.h"
+#include "layer.h"
+#include "snrt.h"
+
+void batchnorm_layer(layer l) {
```
noah95, line 11: the `layer` struct is passed as argument. This is **>200 bytes** of data passed through stack. passing a pointer would make it faster, more energy efficient and reduces the chance of stack-overflow.

sw/applications/src/layers/batchnorm_layer.h
```diff
@@ -0,0 +1,7 @@
+// Copyright 2020 ETH Zurich and University of Bologna.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "layer.h"
+
+void batchnorm_layer(layer l);
```
noah95, line 7: no newline at EOF

sw/applications/src/layers/conv2d_layer.c
```diff
@@ -0,0 +1,396 @@
...
+#include "gemm.h"
+#include "layer.h"
+#include "printf.h"
+#include "snrt.h"
+#include "utils.h"
+
+#define min(a, b) ((a) < (b) ? (a) : (b))
+#define max(a, b) ((a) > (b) ? (a) : (b))
+
+void conv2d_layer(layer l) {
```
noah95, line 16: same: pass by reference

sw/applications/src/layers/conv2d_layer.h
```diff
@@ -0,0 +1,7 @@
+// Copyright 2020 ETH Zurich and University of Bologna.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "layer.h"
+
+void conv2d_layer(layer l);
```
noah95, line 7: docstring?

sw/applications/src/layers/batchnorm_layer.h
```diff
@@ -0,0 +1,7 @@
+// Copyright 2020 ETH Zurich and University of Bologna.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "layer.h"
+
```
noah95, line 6: docstring?

sw/applications/src/layers/maxpool_layer.c
```diff
@@ -0,0 +1,115 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#include "maxpool_layer.h"
+
+#include "layer.h"
+#include "maxpool.h"
+#include "printf.h"
+#include "snrt.h"
+
+void maxpool_layer(layer l) {
```
noah95, line 12: same: pass by reference

sw/applications/src/layers/maxpool_layer.h
```diff
@@ -0,0 +1,7 @@
+// Copyright 2020 ETH Zurich and University of Bologna.
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "layer.h"
+
+void maxpool_layer(layer l);
```
noah95, line 7: docstring

sw/applications/src/net_fusedconv.c
```diff
@@ -0,0 +1,123 @@
...
+#include <math.h>
+
+#include "conv2d.h"
+#include "data_fusedconv.h"
+#include "printf.h"
+#include "snrt.h"
+#include "utils.h"
+
+int main() {
+    float *ptr = (void *)snrt_cluster_memory().start;
```
noah95, line 14: you assume that there is nothing else at TCDM start. There is `snrt_l1alloc` which also performs simple size-checking and ensures proper alignment.

sw/applications/src/net_gemm.c
```diff
@@ -0,0 +1,184 @@
...
+
+    const layer l1_gemm_l = gemm_l;
+
+    volatile uint32_t cluster_num = snrt_cluster_num();
+    volatile uint32_t cluster_id = snrt_cluster_idx();
+    volatile uint32_t compute_num = snrt_cluster_compute_core_num();
+    volatile uint32_t compute_id = snrt_cluster_compute_core_idx();
+
+    void *mat_A, *mat_B, *mat_C;
+    void *ptr = (double *)snrt_cluster_memory().start;
```
noah95, line 30: same: Use `snrt_l1alloc`

sw/applications/src/net_gemm.c
```diff
@@ -0,0 +1,184 @@
...
+    gemm_l.B = (void *)gemm_B_dram;
+    gemm_l.C = (void *)gemm_C_dram;
+
+    const layer l1_gemm_l = gemm_l;
+
+    volatile uint32_t cluster_num = snrt_cluster_num();
+    volatile uint32_t cluster_id = snrt_cluster_idx();
+    volatile uint32_t compute_num = snrt_cluster_compute_core_num();
+    volatile uint32_t compute_id = snrt_cluster_compute_core_idx();
+
```
noah95, line 28: why `volatile`? better `const`

sw/applications/src/net_gemm.c
```diff
@@ -0,0 +1,184 @@
...
+#include "math.h"
+#include "perf_cnt.h"
+#include "printf.h"
+#include "snrt.h"
+#include "utils.h"
+
+#define MAT_ROW_PADDING 4
+#define MAT_PADDING 8
+
+int main() {
```
noah95, line 17: a quick description of what this program does?

sw/applications/src/utils/utils.c
```diff
@@ -0,0 +1,172 @@
+// Copyright 2020 ETH Zurich and University of Bologna.
```
noah95, line 1: duplicate docstring

sw/applications/src/utils/utils.c
```diff
@@ -0,0 +1,172 @@
...
+// Licensed under the Apache License, Version 2.0, see LICENSE for details.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "utils.h"
+
+#include <stdint.h>
+
+#include "../../vendor/riscv-opcodes/encoding.h"
+#include "layer.h"
+#include "math.h"
```
noah95, line 11: cosmetic: if you want the system library, use `<math.h>`

sw/applications/src/utils/utils.h
```diff
@@ -0,0 +1,44 @@
...
+ * to track performance
+ *
+ * @return uint32_t
+ */
+uint32_t benchmark_get_cycle();
+
+/**
+ * @brief start tracking of dma performance region
+ *
+ */
```
noah95, line 19: Perhaps mention that this has no HW side-effects, just for tracking in trace files. Same of course for `stop`

sw/applications/src/utils/utils.c
```diff
@@ -0,0 +1,172 @@
...
+ *
+ * @return uint32_t
+ */
+uint32_t benchmark_get_cycle() { return read_csr(mcycle); }
+
+/**
+ * @brief start tracking of dma performance region
+ *
+ */
+void snrt_dma_start_tracking() { asm volatile("dmstati t0, 1"); }
```
noah95, line 27: clobber `t0` or there will be nasty surprises ;) or use `zero`

sw/applications/src/utils/utils.c
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * @brief stop tracking of dma performance region
+ *
+ */
+void snrt_dma_stop_tracking() { asm volatile("dmstati t0, 3"); }
```
noah95, line 33: clobber `t0` or there will be nasty surprises ;) or use `zero`

noah95 (changes requested):
Thanks for the great work! Please take all my comments with a grain of salt and you choose on what to change.

- Could we add this to the CI? I think it would make sense to protect this valuable contribution from any regressions. Since you already have the data generation and validation in place this shouldn't be too much effort.
- Probably a bit far fetched: But if there were a CI, could we also check the performance and through that protect us from any HW regressions?

sw/applications/include/layer.h
fischeti: yes, that header file is quite outdated and not really necessary. I will change it

sw/applications/src/kernels/conv2d.h
ABurrello: I can definitely update all the calls to layers and to kernels using structs

fischeti: Done.

fischeti:
I also added the applications now to the CI. Only the functionality is checked, not the performance though.

sw/applications/src/kernels/conv2d.h
```diff
@@ -0,0 +1,251 @@
...
+ * @param kappa multiplication factor for BatchNorm
+ * @param lambda bias for BatchNorm
+ * @param flag_relu RELU activation flag
+ * @param flag_batch_norm BatchNorm flag
+ * @param flag_y_accumulate_start indicates that output feature map is
+ * initizialized with zeros
+ * @param flag_y_accumulate_end indicates that BN, RELU can be performed
+ * @param memory_chan Not used
+ */
+void __attribute__((noinline)) occamy_conv_opt_fp64(kernel_fp64 *k);
```
noah95, line 111: Could move this docstring to the struct declaration for better rendering in the docs

noah95 (approved):
Very nice, thanks! Just a small comment on the docstrings, you decide :)

sw/applications/src/kernels/conv2d.h
fischeti: Done.

PR #338, opened 2022-02-14
snRuntime: Optimize function calls via internal linkage

colluca:
Quoting the C11 standard: 
> "Any function with internal linkage can be an inline function. For a function with external linkage, the following restrictions apply: If a function is declared with an inline function specifier, then it shall also be defined in the same translation unit.". 

For performance reasons, inlining of small functions e.g. `snrt_hartid()` is desirable. This is only possible if the function definition is available in the translation unit which desires inlining.
We can achieve this by moving the function definitions into the library header.


noah95 (approved):
Good catch, thanks :+1:

PR #343, opened 2022-02-18
banshee: Support the `fence` instruction

SamuelRiedel:
- Support `fence` instructions with correct ordering for reads and writes. Peripheral ordering is currently not supported.
- Update the MemPool configuration to match the current configuration system's requirements


paulsc96 (approved):
LGTM

PR #344, opened 2022-02-21
banshee: Add zeromemory and bootrom

GiannaP:
**Banshee:**
- ZeroMemory as peripheral
- Configuration: allow to set num-cores, num-clusters, and base-core-id-offset with a configuration file. Commandline parameters given to banshee are ignore if the values are alreadyset by a configuration file.
- added a Bootrom
- new configuration `snitch_cluster.yaml` with zeromemory and bootrom

**SW**:
- differentiating between `snRuntime-cluster`, `snRuntime-banshee`, and `snRuntime-banshee-general`
  - `snRuntime-cluster`: takes `platform/shared/*` and `platform/rtl/*`
  -  `snRuntime-banshee`: takes `platform/shared/*` and `platform/banshee/*`, takes as much of the RTL startup script as possible and does use the Banshee bootrom
  -  `snRuntime-banshee-general`: takes `platform/banshee-general`, original startup code which does not use a bootrom
- restructuring of the various `start_*.S` and `start.c` files according to the three runtime needs

**CI**:
- banshee tests which are general as before using `snRuntime-banshee-general`
- banshee tests which are running on snitch_cluster with the `snitch_cluster.yaml` configuration with `snRuntime-banshee` runtime (e.g. used for zeromemory test which requires a configuration file defining the zeromemory)

**Note/TODOs/Questions:**

- [x] Proabably it would be better when the commandline parameters given to banshee overwrite the configuration file
- [x] delete  `snRuntime-banshee-general`
- [x] delete `start_billywig.S`, `start_thestral.S` files etc.
- [x] This should be rebased and verified on Tim's changes from #337


sw/banshee/src/main.rs
```diff
@@ -215,7 +216,15 @@ fn main() -> Result<()> {
...
-        Configuration::parse(config_file, engine.num_clusters)
+        // if configuration file is given and `architecture` information is set
+        // use that configuration, else banshee parameter
+        let config_used: Configuration = Configuration::parse(config_file, engine.num_clusters);
+        if config_used.architecture.num_clusters != 0 {
+            engine.num_cores = config_used.architecture.num_cores;
+            engine.num_clusters = config_used.architecture.num_clusters;
+            engine.base_hartid = config_used.architecture.base_hartid;
+        }
+        config_used
```
SamuelRiedel, line 227: I agree that it would be nice to have the option to override the config file with command line arguments. Would moving this block above the `num-*` blocks be enough?

sw/banshee/src/engine.rs
```diff
@@ -571,9 +587,12 @@ pub unsafe fn add_llvm_symbols() {
 
 impl CpuState {
     /// Create a new CpuState
-    pub fn new(num_dm: usize) -> Self {
+    pub fn new(num_dm: usize, hartid: usize, bootrom_addr: u32) -> Self {
+        let mut reg_init: [u32; 32] = [0; 32];
+        reg_init[10] = hartid as u32;
+        reg_init[11] = bootrom_addr;
```
SamuelRiedel, line 593: This should come from the config eventually. I am fine with merging this like this for now, but let's create an issue to make sure we make this parameterizable in the future.

SamuelRiedel (commented):
Very nice! The Banshee part looks mostly good to me. One question about the runtimes, now that we have the bootrom, the runtime should be the same for RTL and Banshee, no? I know that the printf might be different, but the boot/init code should work, or are there things missing?

GiannaP:
> Very nice! The Banshee part looks mostly good to me. One question about the runtimes, now that we have the bootrom, the runtime should be the same for RTL and Banshee, no? I know that the printf might be different, but the boot/init code should work, or are there things missing?

I could "merge" a further part. Despite the `put_char.c` there is now mainly the address for the `exit_code` return different. In Banshee this is the first scratch register, in RTL this goes to `tohost` from the fesvr. As I am not too familiar with the fesvr frontend this needs some further investigation to merge the last part.

sw/banshee/src/bootroms.rs
```diff
@@ -0,0 +1,192 @@
...
+            0x0 => self.boot_addr.load(Ordering::SeqCst),
+            0x4 => self.core_count.load(Ordering::SeqCst),
+            0x8 => self.hartid_base.load(Ordering::SeqCst),
+            0xc => self.tcdm_start.load(Ordering::SeqCst),
+            0x10 => self.tcdm_size.load(Ordering::SeqCst),
+            0x14 => self.tcdm_offset.load(Ordering::SeqCst),
+            0x18 => (self.global_mem_start.load(Ordering::SeqCst) & 0xfff) as u32,
+            0x1c => ((self.global_mem_start.load(Ordering::SeqCst) >> 5) & 0xfff) as u32,
+            0x20 => (self.global_mem_end.load(Ordering::SeqCst) & 0xfff) as u32,
+            0x24 => ((self.global_mem_end.load(Ordering::SeqCst) >> 5) & 0xfff) as u32,
```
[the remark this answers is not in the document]
GiannaP: Yes, not sure what I was thinking here...... xD should be fixed now

sw/banshee/src/main.rs
GiannaP: it was not enough, but I implemented it now correctly.

sw/banshee/src/engine.rs
GiannaP: I created an Issue here for now: #350

GiannaP:
Note: The banshee tests in the container are failing because the container needs to be rebuild. This could cause that CI on non-rebased branches (which will use this new container) will fail. Therefore, I will do that once everyone is fine with merging.

sw/banshee/src/bootroms.rs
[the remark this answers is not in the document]
paulsc96: Is it? There is no marked change (yet).

Issue #347, opened 2022-02-24
occamy: PCIE config regbus not connected


fischeti:
`pcie_cfg` is left unconnected. It should be connected to the peripheral crossbar.

paulsc96:
Fixed in #348.

PR #348, opened 2022-02-24
occamy: Add HBI narrow cfg regbus + Connect PCIE cfg regbus

fischeti:
Addresses #347 

## Changes
* Connects `pcie_cfg` to peripheral Xbar which is floating now
* Add new `RegBus` that is used for the narrow Serial Link. This `RegBus` steals the address range of the HBI APB controller which is not in use. The `reg_to_apb` converter was removed
* Refactoring `hbi_cfg` to `hbi_wide_cfg` resp. `hbi_narrow_cfg`


GiannaP (approved):
LGTM! :)

PR #349, opened 2022-02-24
occamy: Preserve ID widths until atomic adapters, add ATOP filters

paulsc96:
* Remove ID serializer in front of SPM
* Move narrow-to-wide ID width conversion behind atomics adapter
* Add ATOP filters in front of all atomics-incapable narrow slaves


GiannaP (approved):
LGTM!

PR #354, opened 2022-03-02
addrmapgen: Make `nr_clusters` configurable

GiannaP:
add support to generate the addressmap (pdf and md) for occamy configurations with a different amount of clusters than 4


paulsc96 (approved):
LGTM

PR #357, opened 2022-03-11
occamy cluster: Add a small fpu to ninth dma core

GiannaP:
The added FPU has only single and double precision support (no SIMD, no Smallfloat, no SDOTP operation group)


paulsc96 (approved):
LGTM

PR #359, opened 2022-03-18
serial link: Attach to narrow xbar + truncate address

fischeti:


paulsc96 (changes requested):
LGTM, but you probably need to regenerate RTL and rebase before we can merge.

paulsc96 (approved):

PR #362, opened 2022-03-21
occamy: Update solder templates for FLL integration

colluca:
Extends the peripherals regbus XBAR to expose one regbus port per FLL at the top level


GiannaP (approved):
LGTM :)

PR #368, opened 2022-03-28
fpu: Parameterize and default-disable FPU IO spills, increase datapath pipelining

suehtamacv:
This PR updates the pipeline count of the FPU datapaths, and sets the `spill_register`s at the boundary of the `snitch_fpu` wrapper module to the bypass mode. This change was motivated by backend runs.


GiannaP:
I was just a bit surprised by the additional pipeline stage of the conversion unit, but I guess the missing spill registers make it necessary. @suehtamacv was it on the critical path or did you preemptively add it?

paulsc96:
> I was just a bit surprised by the additional pipeline stage of the conversion unit, but I guess the missing spill registers make it necessary. @suehtamacv was it on the critical path or did you preemptively add it?

Technically, removing both spill registers and adding one stage on each datapath _reduces_ the total stage count by one. So unless the original design was seriously botched, we should need at least one extra stage :see_no_evil:. But we could investigate this.

PR #126, opened 2021-05-07
occamy/fpga: Prints 'Hello World!' on VCU128

niwis:
The commits of this PR are atomic (might as well be separate PRs).
This PR resolves most concerning critical warnings and timing violations in Vivado, generally optimizes the flow, and adds a primitive bootrom. The key changes are:
* **Add bootrom**: This is a primitive version that just prints _Hello World!_ over the Xilinx AXI UART before parking in the trap vector.
* **Add ILA**: Signals that should be observed can be marked as debug in `vivado_ips/occamy_xilinx_debug.xdc` (internal occamy signals) or `occamy_vcu128_bd.tcl` (BD-level signals), and added to the ILA in `occamy_vcu128_debug.xdc`.
* **Add AXI BRAM**: For bring-up purposes, a BRAM is added and mapped to `0x30000000` of the PCIe AXI.
* **Rework SmartConnect**: Every AXI master has their own SmartConnect, and unassigned slave segments are excluded. This reduces congestion by a lot.
* **Update Address Map**: Peripherals are moved to an uncached region of the PCIe AXI:
    * Xilinx AXI UART: `0x40000000`
    * HBM Stack 0 APB: `0x40400000`
    * HBM Stack 1 APB: `0x40800000`
* **Associate interface clock**: While packaging the `occamy_xilinx` IP, `clk_i` is associated with the AXI interfaces as clock. This also makes the manual `CONFIG.FREQ_HZ` override in `occamy_vcu128_bd.tcl` superfluous.
* **Create dedicated real time clock**: Add a new clock (25MHz, phase 90 deg.) to the clock wizard to serve as real-time clock for occamy. This resolves a hold violation.

PR #123, opened 2021-05-03
OpenSBI booting in sim

zarubaf:
This PR adds the necessary capabilities to boot OpenSBI in simulation for Occamy. Right now the `generic` platform works as most of the stuff is figured out by looking at the device tree. The log from the simulation:

```
OpenSBI v0.9
   ____                    _____ ____ _____
  / __ \                  / ____|  _ \_   _|
 | |  | |_ __   ___ _ __ | (___ | |_) || |
 | |  | | '_ \ / _ \ '_ \ \___ \|  _ < | |
 | |__| | |_) |  __/ | | |____) | |_) || |_
  \____/| .__/ \___|_| |_|_____/|____/_____|
        | |
        |_|

Platform Name             : eth,occamy
Platform Features         : mfdeleg
Platform HART Count       : 1
Platform IPI Device       : clint
Platform Timer Device     : clint
Platform Console Device   : htif
Platform HSM Device       : ---
Platform SysReset Device  : htif
Firmware Base             : 0x80000000
Firmware Size             : 104 KB
Runtime SBI Version       : 0.3

Domain0 Name              : root
Domain0 Boot HART         : 0
Domain0 HARTs             : 0*
Domain0 Region00          : 0x0000000004000000-0x000000000400ffff (I)
Domain0 Region01          : 0x0000000080000000-0x000000008001ffff ()
Domain0 Region02          : 0x0000000000000000-0xffffffffffffffff (R,W,X)
Domain0 Next Address      : 0x0000000080200000
Domain0 Next Arg1         : 0x0000000082200000
Domain0 Next Mode         : S-mode
Domain0 SysReset          : yes

Boot HART ID              : 0
Boot HART Domain          : root
Boot HART ISA             : rv64imafdcsu
Boot HART Features        : scounteren,mcounteren
Boot HART PMP Count       : 16
Boot HART PMP Granularity : 8
Boot HART PMP Address Bits: 54
Boot HART MHPM Count      : 0
Boot HART MHPM Count      : 0
Boot HART MIDELEG         : 0x0000000000000222
Boot HART MEDELEG         : 0x000000000000b109
```

PR #370, opened 2022-03-30
occamy: Update `riscv_dbg` to version with fixed reset issues

paulsc96:
* [x] backport `common_cells` changes from [#126](https://github.com/pulp-platform/common_cells/pull/126) as a dependency.
* [x] backport `riscv_dbg` fixes from [#123](https://github.com/pulp-platform/common_cells/pull/126) as fixes of interest.

### DIFF
--- a/hw/vendor/patches/pulp_platform_common_cells/0006-common_cells-Clearable-cdcs-126.patch
+++ b/hw/vendor/patches/pulp_platform_common_cells/0006-common_cells-Clearable-cdcs-126.patch
@@ -1,0 +1,2557 @@
+From 41f8cdc18733b8f57e0f81a30650e4093b4f3d76 Mon Sep 17 00:00:00 2001
+From: Paul Scheffler <paulsc@iis.ee.ethz.ch>
+Date: Wed, 30 Mar 2022 21:13:57 +0200
+Subject: [PATCH] common_cells: Clearable cdcs (#126)
+
+---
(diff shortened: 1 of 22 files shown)


meggiman:
Since the common_cells PR [126](https://github.com/pulp-platform/common_cells/pull/126) finally landed it should not be long before we can merge the upstream [PR](https://github.com/pulp-platform/riscv-dbg/pull/123). I would suggest we wait until this happens and vendor in the new version. The current patch seems to be missing some modifications (e.g. the TB modifications to verify the new reset behavior). Have you already been able to simulate the entire design with the modified IP?

paulsc96:
> Since the common_cells PR [126](https://github.com/pulp-platform/common_cells/pull/126) finally landed it should not be long before we can merge the upstream [PR](https://github.com/pulp-platform/riscv-dbg/pull/123). I would suggest we wait until this happens and vendor in the new version. 

Can do. I opened the branch primarily so we can review the changes and verify that this resolves our issues.

> The current patch seems to be missing some modifications (e.g. the TB modifications to verify the new reset behavior). 

The way we vendored this repo excludes testbenches, since we do not need them. I dislike the use of vendor for our own dependencies (see #371) but it's a bad time to rock the boat too much, so let's keep things simple for now.

> Have you already been able to simulate the entire design with the modified IP?

@thommythomaso will look into this tomorrow.

PR #372, opened 2022-04-04
cluster: Parameterize debug port, remove and internally tie static ports

paulsc96:
* Add a configuration parameter `enable_debug` tying off the cluster debug request port for optimization.
* Remove `debug_req_i` port if `enable_debug` is false.
* Remove `clk_d2_bypass_i` port if iso-crossings disabled (no longer determined by `tie_ports`).
* Propagate port changes through Occamy.


GiannaP (approved):
LGTM!

PR #375, opened 2022-04-08
Upgrade AXI to master

GiannaP:
In this PR the AXI is upgraded to the latest stable master commit.

**Done:**
- moved DMA from AXI branch into `hw/ip/future`
- upgraded AXI to master branch
- went through old patches and checked that they were indeed merged and not needed anymore.
- added three new patches:
  - rename generated axi types `*_resp_t` to `*_rsp_t` to avoid solder adaptations
  - vcs compatibility fix for `axi_isolate`: change unsigned parameters to signed 
  - vcs compatibility fix for `axi_demux`: split up spill registers

**TODO:**
[x] reqrsp_interface: standalone TB fails. Investigation: TB was never adapted for some AMO fixes (see new issue #378)


thommythomaso (commented):
In the file: `hw/ip/future/src/dma/axi_dma_data_mover.sv` the fix preventing a buffer overflow is missing. @fischeti is the fix not yet incorporated here?

fischeti:
> In the file: `hw/ip/future/src/dma/axi_dma_data_mover.sv` the fix preventing a buffer overflow is missing. @fischeti is the fix not yet incorporated here?

No, I merged the fix in a later PR #384 which is already merged. So rebasing should do the job

PR #379, opened 2022-04-12
riscv-dbg: Re-align `dmi_bscane_tap` interface with `dmi_jtag_tap`

noah95:
The current `occamy` design does not synthesize for FPGA because of a pin-mismatch between `dmi_jtag_tap.sv` and `dmi_bscane_tap.sv` that was introduced in #373.

I also contributed this [upstream](https://github.com/pulp-platform/riscv-dbg/pull/130).

CC: @bluewww @meggiman

Edit: Removed the patch and insted bumped to `v0.5.1` that includes this patch


bluewww:
Merged in upstream and released as https://github.com/pulp-platform/riscv-dbg/releases/tag/v0.5.1

paulsc96 (approved):
Excellent, thanks :+1:

PR #382, opened 2022-04-13
occamy: add a system-level DMA

thommythomaso:
Add a system-level DMA connected to the 512bit wide SoC interconnect. The DMA features are:
- A register-based frontend
- Configurable by all cores as well as any debug access

It will mainly be used to:
- Initialize HBM
- Copy between HBM channels
- Copy between the HBM and the system-level SPM


GiannaP (commented):
Just one quick question: I assume the  `hw/ip/future/src/idmareg64_frontend_reg_*.sv` files architecture independent / not-occamy specific (also as you do not add the `.hjson` file? Otherwise, one could move them together with an `.hjson` into its own directory under `hw/system/occamy/src/idma` as done e.g. for [occamy_quadrant_s1_ctrl](https://github.com/pulp-platform/snitch/tree/master/hw/system/occamy/src/occamy_quadrant_s1_ctrl) or [clint](https://github.com/pulp-platform/snitch/tree/master/hw/system/occamy/src/clint).

Otherwise it looks good to me :)

PR #384, opened 2022-04-14
occamy: DMA & AXI transaction tuning

fischeti:
## DMA
* Increases Size of `last_twod_buffer` queue which did not account for all the transactions buffered inside the AXI interconnect
* Adds patch for generating backpressure if there are too many AXI transactions in flight

## AXI tuning
The number of outstanding AXI transactions is increased across the entire system and was verified with the DORY runs we currently have. The number of outstanding transactions increases in each level of hierarchy:
* **Quadrant:** RO-cache, Narrow & Wide Xbar -> 32
* **SoC:** Quadrant-Pre-Xbar , Wide-Xbar, Quadrant-Inter-Xbar -> 64
* **HBM:** HBM-Xbar (+ Simulation memory) -> 128


paulsc96:
As this probably relates to #376: Currently, the quadrant TLBs support 32 outstanding transactions. This should then be sufficient / in line with this configuration and not bottleneck throughput, right? Maybe we should reverify this once the TLBs are merged.

fischeti:
Yes, that should be in line. I can wait for #376 to verify it again👍

paulsc96:
> Yes, that should be in line. I can wait for #376 to verify it again+1

Re-verification would be much appreciated, but we should go ahead and merge this first anyways once reviewed. Tuning the TLB should be part of the TLB MR.

paulsc96 (approved):
I am somehow not an automatic reviewer, but here comes my approval anyways :airplane:

PR #386, opened 2022-04-19
add Wide SPM and Zero Memory

GiannaP:
- add 8GiB of ZeroMemory on wide AXI Xbar
- add 2MiB of 512-bit SPM on wide AXI Xbar


paulsc96 (approved):
LGTM overall. Just be sure to double-check all places interrupts are hardcoded (don't we need to adapt the device tree?) and to add the needed implementation control registers for the new SPM at the chip level later.

PR #395, opened 2022-04-27
cluster: Extend metrics, add performance counters

fischeti:
* Increases number of performance counters from 2 to 16 (8 for core local metrics and 8 for cluster specific metrics e.g. DMA)
* Adds core local retired instruction metrics to cluster performance counter to have everything in one place


hw/ip/snitch/src/snitch.sv
```diff
@@ -313,6 +315,10 @@ module snitch import snitch_pkg::*; import riscv_instr::*; #(
   logic [63:0] instret_q;
   `FFAR(cycle_q, cycle_q + 1, '0, clk_i, rst_i)
   `FFLAR(instret_q, instret_q + 1, !stall, '0, clk_i, rst_i)
+  assign core_events_o.retired_instr = !stall;
```
paulsc96, line 318: could we register these? at least for the `stall` signal, it was highly critical before TLB removal, and probably still is. So unless a core event is already (close to) a register output, we should be careful here.

In any case, we already have retired instruction counter CSRs in the core IIRC; maybe you want to check which of these events are redundant.

fischeti: Yes, I will use register for them

And yes, some of them are redundant. The idea was to have all the performance counters in one place for convenience's sake. Because we only record the events and not the values itself this should hopefully not cause a large overhead

hw/ip/snitch/src/snitch.sv
```diff
@@ -311,8 +313,20 @@ module snitch import snitch_pkg::*; import riscv_instr::*; #(
...
   `FFAR(cycle_q, cycle_q + 1, '0, clk_i, rst_i)
   `FFLAR(instret_q, instret_q + 1, !stall, '0, clk_i, rst_i)
+  `FFAR(retired_instr_q, !stall, '0, clk_i, rst_i)
+  `FFAR(retired_load_q, retire_load, '0, clk_i, rst_i)
+  `FFAR(retired_i_q, retire_i, '0, clk_i, rst_i)
+  `FFAR(retired_acc_q, retire_acc, '0, clk_i, rst_i)
+  assign core_events_o.retired_instr = retired_instr_q;
+  assign core_events_o.retired_load = retired_load_q;
+  assign core_events_o.retired_i = retired_i_q;
+  assign core_events_o.retired_acc = retired_acc_q;
```
paulsc96, line 329: Could we have an else block here tying the additional core events?

paulsc96 (approved):
We can merge this; just one caveat below you may want to consider before that.

PR #397, opened 2022-04-28
axi-riscv-atomics: Update to latest version

SamuelRiedel:
This PR updates the ATUN to the latest version.

- Guarantees atomicity by tracking downstream writes (limits number of writes in-flight and needs to be configured according to the performance requirements)
- Adds the option to use the USER signal to distinguish different harts. This is currently still disabled.

To make it work with VCS, we need to following two updates to dependencies:
- [x] https://github.com/pulp-platform/common_cells/pull/138
- [x] https://github.com/pulp-platform/axi_riscv_atomics/pull/19

PR #399, opened 2022-04-28
occamy: Transfer hartid via user signal

niwis:
This PR
- adds a user width converter to solder,
- increases the user width of the narrow xbar from 1 to 9,
- adds user width converters between the narrow and wide xbar,
- sends `hartid + 1` of all cores via the AXI user signal with each data request (`user = 0` is invalid).

ToDo:
- [x] Test
- [x] Fix linting
- [x] Configure amo adapter to use user signal (requires #397)


niwis:
Ready to merge (if CI passes)

PR #401, opened 2022-04-30
snitch: Fix LR/SC responses

SamuelRiedel:
Currently, all interfaces that return store-conditional responses simply set the response data to 1. However, most memory interfaces are wider than 32 bits and, depending on the alignment, Snitch only sees a slice of the returned value. The result is that store-conditionals not aligned to the interface width are reported as successful, independent of their actual outcome.

This PR fixes the modules that issue/convert the responses by returning -1 instead of one to guarantee that a non-zero return code reaches Snitch independent of any alignment issues. The RISC-V spec specifies a zero return value on success and a non-zero on failure, so we are free to return -1.

I fixed the:
- `reqrsp_interface`, which handles the AXI interface
- `tcdm_amo_shim`, which handles the atomics for the TCDM banks

Are there any other modules that resolve atomics in Snitch or that convert between AXI and Snitch?

I also adapted @colluca's test to check those memory regions. We currently don't run it on Banshee. Banshee replaces the SC with a CAS where it uses the last value loaded by an LR as the comparison value. This synthetic test currently leads to SCs succeeding because we do the same SC/LR/SC test pattern on different addresses. I will fix this later, but the HW fix should be merged as soon as possible.

### DIFF
--- a/hw/ip/reqrsp_interface/src/reqrsp_to_axi.sv
+++ b/hw/ip/reqrsp_interface/src/reqrsp_to_axi.sv
@@ -257,7 +257,9 @@ module reqrsp_to_axi import reqrsp_pkg::*; #(
     // store conditional for the reqrsp interface.
     if ((axi_rsp_i.b_valid && ~atomic_in_flight_q
           && axi_rsp_i.b.resp != axi_pkg::RESP_EXOKAY)) begin
-      reqrsp_rsp_o.p.data = 1;
+      // Set all 32-bit words to 1 since we don't know the alignment
+      // and which bits are cut off by the core.
+      reqrsp_rsp_o.p.data = {DataWidth/32{32'h1}};
     end
   end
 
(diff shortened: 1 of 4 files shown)


colluca:
I have yet to check the HW changes but from the description of the PR I do see a fix to be made. Unfortunately I fell into the same mistake when I first created this [PR](https://github.com/openhwgroup/cva6/pull/838) for CVA6.

While the spec says the return code can be any non-zero value, if you read just a bit further it also states:
> The failure code with value 1 is reserved to encode an unspecified failure. Other failure codes are
reserved at this time, and portable software should only assume the failure code will be non-zero.

I had actually tried it out the same way you did, i.e. returning -1, but the RISC-V tests would then fail, as they explicitly check for a return value strictly equal to 1. I proposed to loosen this check in the riscv-tests repo with this [PR](https://github.com/riscv-software-src/riscv-tests/pull/376) where Andrew Waterman in person corrects me pointing out the requirement above.

SamuelRiedel:
Thanks for this hint. I updated the HW changes accordingly and now return a 1 in every 32-bit slice.

PR #420, opened 2022-06-02
sssr: Fix an inferrable timing arc through streamctl FIFO

paulsc96:


paulsc96:
Closed, since the suspicion of a timing loop arose from incorrect elaboration.

PR #429, opened 2022-08-10
Add DRAM preloading in banshee

viv-eth:
Add these flags to the banshee arguments: 

1. `--file-paths=<path-to-binaries>`  
2. `--mem-offsets=<memory-offsets>` 

The file paths are the paths to your dataset data, which should be in binary format. If you have multiple data files you can separate them by commas. In order to save them at a specific location of the main memory make sure you enter a start address (`mem_offset`) for every binary file. As before, it can be a comma-separated list. The user has to be aware of the memory layout and make sure that the offsets are big enough so that data does not get overwritten. There is currently no warnings supported if memory gets overwritten in DRAM.

### DIFF
--- a/sw/banshee/src/main.rs
+++ b/sw/banshee/src/main.rs
@@ -308,7 +308,7 @@ fn main() -> Result<()> {
 
         trace!("Train data starts at address: 0x{:x}", mem_offset);
 
-        let train_data = dram_preload::bin_read(bin_path, mem_offset).unwrap();
+        let train_data = dram_preload::generic_bin_read::<4>(bin_path, mem_offset).unwrap();
 
         let train_data_length = train_data.len() as u64;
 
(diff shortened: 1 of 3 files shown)


sw/banshee/src/main.rs
```diff
@@ -137,6 +144,30 @@ fn main() -> Result<()> {
...
+                .long("train-labels-bin-file-path")
+                .takes_value(true)
+                .help("Path to the binary file of the labels for training."),
+        )
+        .arg(
+            Arg::with_name("train-labels-mem-offset")
+                .long("train-labels-mem-offset")
+                .takes_value(true)
+                .help("Define the offset at which the train labels should be written into DRAM."),
+        )
```
paulsc96, line 170: Can we maybe generalize this to not be specific to training data and labels? We could provide a list of preloadable binary files in the style `--bin-files <file>:<addr>,<file>:<addr>,..` or something along these lines.

sw/banshee/src/main.rs
```diff
@@ -260,6 +293,57 @@ fn main() -> Result<()> {
...
+        mem.extend(train_data);
+
+        for addr in mem_offset..mem_offset + train_data_length {
+            let val: u32 = mem.get(&(addr)).copied().unwrap_or(0);
+            trace!("address = 0x{:x}, binary value = {:#034b}", addr, val);
+        }
+    }
+
+    if has_train_labels_bin {
+        let bin_path = matches.value_of("train-labels-bin-file-path").unwrap();
```
paulsc96, line 326: We can get rid of this duplication if we treat preloadable files as a generic list as proposed above. Question is how to treat the different data sizes; they could also be passed similar to the preloading address if necessary.

paulsc96 (changes requested):
Thanks for the contribution :)

I think this one requires a bit of rethinking. I am not entirely convinced we need a preloading mechanism outside of ELF sections (i.e. compiling / patching your binary to specify additional sections). But assuming that we do, we should do it in a more generic fashion; see below.

PR #432, opened 2022-08-29
banshee: Disable opaque pointers in JIT code

SamuelRiedel:
The nightly version of Rust seems to enable opaque pointers per default. However, our current `llvm-sys` setup does not yet support those types. Since we need to fuse the IR generated by the Rust compiler with our own, generated with llvm-sys, we have to disable opaque pointers for the Rust compiler manually.

Simultaneously this fixes two minor style warnings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ jobs:
         - stable
         - beta
         - nightly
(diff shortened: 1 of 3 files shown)


paulsc96 (approved):
LGTM :+1:

PR #444, opened 2022-10-28
docs: Add Xfrep extension and publications section

colluca:


colluca:
@paulsc96 There is one last thing I don't understand about the FREP instructions, and it's the `stream_ctl` field which is not reported in the original paper. Do you know what its purpose is? I could add it to the documentation and then it should be complete.

mkdocs.yml
```diff
@@ -8,8 +8,8 @@ theme:
   icon:
     repo: fontawesome/brands/github
 
-repo_url: https://github.com/pulp-platform/snitch
-repo_name: pulp-platform/snitch
+repo_url: https://github.com/colluca/snitch
```
paulsc96, line 11: Let's revert this to its original state

paulsc96 (changes requested):
I think that if we add a dedicated publication section, we should also update the main README accordingly (see bottom).

mkdocs.yml
colluca: Good catch, thanks!

colluca:
> I think that if we add a dedicated publication section, we should also update the main README accordingly (see bottom).

This is my proposal, what do you think? I think it is important to emphasize that anyone using the core should cite the Snitch paper, whereas the other works are more specific. I would maybe move the SSR paper also below the separator line.

paulsc96 (approved):

PR #459, opened 2023-03-01
tracer: Correct accelerator writeback bug

colluca:
When examining the traces, the wrong register and data is specified for writeback from the accelerator interface. E.g.:
```
           6615000    0x80000a74 dmstati t0, 2                  #; 
           6618000                                              #; (acc) t4  <-- 0xfe029ee3
```
Which should in this case instead be:
```
           6615000    0x80000a74 dmstati t0, 2                  #; 
           6618000                                              #; (acc) t0  <-- 0
```

Indeed the tracer RTL dumps the wrong signals. This PR fixes this bug.


paulsc96 (approved):
LGTM

PR #462, opened 2023-03-17
cva6: Tracer prints wrong rs register for fsd instruction

colluca:
Patches the issue described in https://github.com/openhwgroup/cva6/issues/1118 with the fix in https://github.com/openhwgroup/cva6/pull/1123.


niwis:
yes indeed, this will be in the bumped version.

colluca:
This will be integrated with #464.

PR #465, opened 2023-03-22
riscv-tests: Add riscv-tests with vendor

SamuelRiedel:
Vendor in the `riscv-test` to enable instruction based verification.


paulsc96 (approved):
LGTM

PR #478, opened 2023-03-24
`cva6`: Bump to `pulp-v0.1.1`

niwis:
Bump CVA6 to `pulp-v0.1.1`. This does not include any RTL changes but rebases CVA6 on the new `pulp-v0.1.1` tag, which incorporates some of the previous patches. The remaining patches are cleaned up.


paulsc96 (approved):
Looks good to me :)

PR #481, opened 2023-04-04
banshee: changes necessary for automatic multi cluster config

viv-eth:
The structure of the configuration files changes. 
Now you can define the `offset` and `start` of each memory region and it will be automatically computed in the background by `banshee`. C.f. `multi_cluster_simple.yml` for an example.


SamuelRiedel (approved):
The Banshee part looks good to me and seems to pass the CI. That the container test fails is expected since it will have the previous version of Banshee, which will not be compatible with the new test.

I will take those changes over to the new Banshee repository.

paulsc96:
Closing this under the assumption this has been integrated in the new banshee repository.

PR #486, opened 2023-04-19
ci: Update Ubuntu image and replace some actions with PULP defaults

colluca:


GiannaP:
Can we un-draft this PR and merge it or are there any roadblockers?

colluca:
I believe we can yes

PR #500, opened 2023-09-18
Remove combinational loop in axi_dma_data_path.sv

uge:
In the axi_dma_data_path.sv module there is a always_comb block which has a feedback look in out_mask which results in a zero time simulation infinite loop with xsim. This PR restructures the logic to avoid feedback.

### DIFF
--- a/hw/ip/future/src/dma/axi_dma_data_path.sv
+++ b/hw/ip/future/src/dma/axi_dma_data_path.sv
@@ -209,13 +209,11 @@ module axi_dma_data_path #(
   logic                 is_last_w;
 
   always_comb begin : proc_out_mask_generator
-    // default case: all ones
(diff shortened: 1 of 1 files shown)


paulsc96 (commented):
I don't like the way this is written, but technically this is correct. 

These are blocking assignments, so the evaluation of a single block invocation is correct. One invocation can also not trigger another through `out_mask`; refer to IEEE 1800-2017 9.2.2.1 b), which explicitly states that expressions written to in an `always_comb` block are not part of its implicit sensitivity list even when read.

So this seems to be a bug in xsim. That said, I think we should adopt this as it is cleaner.

Since this repo is being archived, I will close this PR and reference it in the successor repo: https://github.com/pulp-platform/snitch_cluster

